PR #28, opened 2017-02-07
Support Map

tafia:
map will correspond to `HashMap` in rust standards


tafia:
I removed several functions from BytesReader, which means this PR is a breaking change

tafia:
This PR was much bigger than anticipated.
I fixed some small issues and refactor the codegen (pb-rs) quite a lot.

Issue #23, opened 2017-02-02
Perf


tafia:
A probably easy perf boost could be done for packed fields with wire_type 1 or 5 (fixed size) ... 
We could return for instance a `Cow<[f64]>` instead of a `Vec<f64>`

PR #34, opened 2017-02-13
Enhance fixed size packed fields

tafia:
When a field has a fixed size, is repeated and packed, we can directly read its data as a slice of data.
So instead of reading each item one by one we convert the entire slice at once. On top of that, we do not need to allocate data and can work with Cow<[M]> instead.

This has a much better perf:

```
 name                         master ns/iter  fixed ns/iter  diff ns/iter   diff %
 read_repeated_packed_float   19,008          277                 -18,731  -98.54%
 write_repeated_packed_float  9,619           2,841                -6,778  -70.46%
```


tafia:
Closes #23

PR #43, opened 2017-02-14
encapsulate oneofs into mod_Message module

### LINKED ISSUE #39: generated oneof structs must be unique
I have some proto definitions which use `oneof` but re-use the same name within different messages. So the generated structs need to be namespaced in some way, ideally with name of enclosing message.

tafia:
Closes #39 

Like for nested enums and messages, oneofs are now encapsulated in `mod_MessageName` module.

@stevedonovan, is it ok for you


stevedonovan:
That does look like a clean solution!

PR #45, opened 2017-02-15
Sanitize rust keywords conflicts

### LINKED ISSUE #41: Rust keywords not escaped as field names
Obvious one, I suppose. People are particularly fond of using the word `type` it seems! Maybe render as `f_type` or something like that?

tafia:
Closes #41 

Uses a static list of keywords (stolen from rust-protobuf) from which I've added all struct/enums used in codegen.


tafia:
@stevedonovan, I'm merging it. I guess it fixes your issue for `type` fields.

PR #49, opened 2017-02-15
errors: add Into<::std::io::Error> conversion

koivunej:
This conversion helps when working with for example tokio-proto where
`Codec` must use `std::io::Error`.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -34,3 +34,16 @@ error_chain! {
         }
     }
 }
+
+impl Into<::std::io::Error> for Error {
+    fn into(self) -> ::std::io::Error {
+        use ::std::io;
+        match self {
+            Error(ErrorKind::Io(x), _) => x,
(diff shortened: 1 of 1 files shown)


tafia:
Thanks for the PR!
We could probably git rid of all these io compatible variants directly when parsing.
There is no real reason for having them apart me not knowing io errors well.
Could you make the change before merging else I'll happily do it next.

koivunej:
The only one of those I'd like to get rid of is `std::string::FromUtf8Error` which is essentially a wrapper around `std::string::Utf8Error` and the input that was incorrent. Luckily it was not used anywhere.

I think there is still value in keeping the variants for unit testing and anyone using the library in more permissive environment. I'll post another commit with the extra error removed.

koivunej:
Looking at the diffs again, you probably meant removing of `errors::ErrorKind::Eof` in favor of the `std::io::ErrorKind::UnexpectedEof`. That's now removed as well.

tafia:
Thanks!!

Issue #27, opened 2017-02-06
Add more options for pb-rs


tafia:
Example:
- generate read-only, read-write or write-only
- generate `Cow` or owned version
- nested types and import use `mod_Message` or eventually a rust-protobuf like solution in replacing all hierarchical calls `a.b.c `into `a_b_c`
- be verbose
- work on a directory
- change destination filename

Issue #48, opened 2017-02-15
Problem with Multiple Proto Files


stevedonovan:
Looking good so far, but there's a hurdle to be jumped: when multiple proto files are defined, the lookup is ambiguous. Consider `package foo.one`  (by convention the filename would be `one.proto` but not necessarily.)  Say there's also packages `foo.two` and foo.three` and `foo.one` refers to messages in them. They are in a directory `test` - the full absolute crate path of anything in package `foo.one` is `::test::foo::mod_foo::mod_one`. For references in `one.rs` the strategy is to bring in `mod_foo` from `two` and `mod_foo` from `three`.  But `mod_foo` is a _nested_ module of `::test::foo::two` and a _different_ nested module of `::test::foo::three`. Hence the ambiguity!

One solution is always to emit full paths, like `::test::two::mod_foo::mod_two::Two`, and that would work. But the paths are starting to get ugly for human programmers!  So another solution is not to create those nested `mod_NAME` modules and instead rely on the usual rules. That is, the .rs files are created in a directory `foo`, and there's a generated `mod.rs` that refers to them. The full path would now be simply `::foo::two::Two`. (Without the `mod_` but with usual escaping rules for keywords) That's simple enough to be used directly in generated code.

I'm happy to work on a PR but this is a disruptive change, so I thought that some discussion is needed on the way forward.

tafia:
Thanks for writing this. I too found the packages modules not nice to work with. 

Just to be sure we're on the same page you propose to ditch the packages inner modules and rely **only** on rust files modules, right? So the `::foo::two::Two` would become, if we follow the previous example, more a `::test::two::Two` (as there is no `foo` in filename) ?

Having full paths everywhere (the simplified version) does fix the ambiguity issue but I'm not sure it is the best solution as it makes refactoring very painful. Whenever you want to change your file hierarchy you need to regenerate all the files (provided pb-rs give you an option to prefix all these absolute paths), This can be inconvenient and problematic if you are not very careful that you're using exactly the same version of pb-rs,

So if possible, I'd like having relative paths using `super::` which still follow the usual rule: instead of using `Two` of `::test::two::Two` we could use `super::two::Two`.

The dev may not be that straightforward but if you feel brave enough I'll appreciate that!
The library is very young so I don't think this wil be too disruptive,

Do you see any issue in using the `super::` version?

stevedonovan:
[...]

PR #50, opened 2017-02-16
Introduce clap for pb-rs cmdline parsing

koivunej:
Related to issue #27, this PR lays out the foundation for more options with the use of [clap-rs](https://github.com/kbknapp/clap-rs). There is one new option already implemented (--single-mod or -s). While probably poorly named it does what I hoped it to do: I do not wish there to be modules mod_A.mod_B.mod_C as the source of my .proto file has.

One thing that worries me is testing, which should probably be thougth out before adding any more options. My current implementation of the option is probably too short sighted at the moment, even though it seems to WorksForMe&copy; :)

In addition to testing, you'll probably notice that I used overly long lines to describe the options. I'd wish `rustfmt` was used so that there would be no worrying about the formatting :) However, as I tried to use it I ran into [rustfmt#1265](https://github.com/rust-lang-nursery/rustfmt/issues/1265) which makes using it quite unbearable in the presence of generated code files. For example: time to process a single generated `examples/codegen/data_types.rs` is 1m30s which can be easily optimized to 1min, but that's still way too long.

### DIFF
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -2,40 +2,36 @@
 extern crate nom;
 #[macro_use]
 extern crate error_chain;
+#[macro_use]
+extern crate clap;
 
 mod parser;
 mod types;
 mod errors;
 mod keywords;
 
-use std::env;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use clap::{App, Arg};
 use types::FileDescriptor;
 
 fn main() {
+    let matches = App::new("pb-rs").version(crate_version!())
+        .arg(Arg::with_name("INPUT").required(true).index(1).help("The .proto file used to generate quick-protobuf code").validator(|x| ends_with_proto(x)))
+        .get_matches();
 
-    let args = env::args().collect::<Vec<_>>();
-    let usage = format!("{} <file.proto>", args[0]);
-
-    if args.len() == 0 {
-        println!("{}", usage);
-        return;
-    }
-
-    let in_file: PathBuf = args.get(1).map(|a| a.into()).unwrap();
-    match in_file.extension().and_then(|e| e.to_str()) {
-        Some("proto") => (),
-        _ => {
-            println!("{}", usage);
-            println!("\r\nExpecting an input file with 'proto' extension");
-            return;
-        }
-    }
-
+    let in_file: PathBuf = matches.value_of("INPUT").map(|a| a.to_string().into()).unwrap();
     let out_file = in_file.with_extension("rs");
 
     FileDescriptor::write_proto(&in_file, &out_file)
         .expect(&format!("Could not convert {} into {}", 
                          in_file.display(), out_file.display()));
 
+}
+
+fn ends_with_proto<P: AsRef<Path>>(path: P) -> Result<(), String> {
+    match path.as_ref().extension() {
+        Some(x) if x == "proto" => Ok(()),
+        Some(x) => Err(format!("Expected path with extension 'proto', not: '{}'", x.to_string_lossy())),
+        None => Err(format!("Expected path with extension 'proto'")),
+    }
 }
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -12,20 +12,27 @@ mod keywords;
 
 use std::path::{Path, PathBuf};
 use clap::{App, Arg};
-use types::FileDescriptor;
+use types::{FileDescriptor, Config};
 
 fn main() {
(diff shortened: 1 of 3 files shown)


codegen/src/main.rs
```diff
@@ -2,40 +2,44 @@
...
-use std::env;
-use std::path::PathBuf;
-use types::FileDescriptor;
+use std::path::{Path, PathBuf};
+use clap::{App, Arg};
+use types::{FileDescriptor, Config};
 
 fn main() {
+    let matches = App::new("pb-rs").version(crate_version!())
+        .arg(Arg::with_name("OUTPUT").required(false).long("output").short("o").takes_value(true).help("Generated file name, defaults to INPUT with 'rs' extension").validator(|x| extension_matches(x, "rs")))
```
tafia: Could you add some line breaks?
I don't really like horizontally scrolling

codegen/src/types.rs
```diff
@@ -1046,31 +1052,35 @@ pub struct FileDescriptor {
 
 impl FileDescriptor {
 
-    pub fn write_proto<P: AsRef<Path>>(in_file: P, out_file: P) -> Result<()> {
-        let mut desc = FileDescriptor::read_proto(&in_file)?;
-        desc.fetch_imports(in_file.as_ref())?;
+    pub fn write_proto<P: AsRef<Path>>(config: &Config<P>) -> Result<()> {
```
tafia: I think this fn could go in `Config` instead. 
Then we could probably have `FileDescriptor` private.

tafia:
Thanks!
cc @stevedonovan as we spoke about teaching pb-rs how to work with many files.

I'm not very familiar with clap (I use to work with docopt instead) but I like what you've done.
I made some small comments, nothing major. 

This is a good start for #27.

tafia:
> I do not wish there to be modules mod_A.mod_B.mod_C as the source of my .proto file has

I believe @stevedonovan is working on it already (see #48). We may probably get rid of them entirely.

> One thing that worries me is testing, which should probably be thougth out before adding any more options. 

Yes. Today, the script on appveyor will regenerate all examples and test files before running the tests. This is not perfect but it works. We could probably have some script which test different inputs in the future.

> For example: time to process a single generated examples/codegen/data_types.rs is 1m30s which can be easily optimized to 1min, but that's still way too long.

I haven't tried it yet. We could add a flag to skip rustfmt and clippy warning like [what rust-protobuf does](https://github.com/stepancheg/rust-protobuf/blob/master/src/lib/descriptor.rs#L4-L8)

koivunej:
> We could add a flag to skip rustfmt and clippy warning like what rust-protobuf does

I got some WIP here: [https://github.com/koivunej/quick-protobuf/tree/rustfmt-ignore]. No clippy stuff yet though.

codegen/src/types.rs
koivunej: Design-wise I'd prefer `Config` to be a simple struct with no additional functionality outside of mangling the configuration values (perhaps some of the path mangling in `write_proto` could move there). Assuming later on with more options, it will be borrowed deeper and deeper into the methods/functions doing the actual job. While I do not yet see much reason for making `FileDescriptor` private perhaps it could be managed with a free function like below instead?

```
pub fn write_proto<P: AsRef<Path>>(config: &Config<P>) -> Result<()> {
  // ...
}
```

koivunej:
The proposed output of `cargo run -- --help` (with all the metadata from `Cargo.toml`):

```
pb-rs 0.1.0
Johann Tuffe
A converter from proto files into quick-protobuf compatible rust module
USAGE:
    pb-rs [FLAGS] [OPTIONS] <INPUT>
FLAGS:
    -s, --single-mod    Omit generation of modules for each package when there is only one package
    -h, --help          Prints help information
    -V, --version       Prints version information
OPTIONS:
    -o, --output <OUTPUT>    Generated file name, defaults to INPUT with 'rs' extension
ARGS:
    <INPUT>    The .proto file used to generate quick-protobuf code
```

tafia:
I prefer it like this, thanks.
For the `FileDescriptor` I think it would be better to have it private but there is no hurry at all.
I'll let some time to see if @stevedonovan in particular wants to change something then I'll merge (probably tomorrow morning HKT).

stevedonovan:
I'm cool - it's a good idea to have command line parsing - then we can have flags controlling output, etc, and sensible errors. So go ahead and in the meantime I'll plot my next move ;)

tafia:
Thanks @koivunej this will really helps in the future

tafia:
Just published v0.5.0

PR #51, opened 2017-02-16
Rustfmt and clippy ignores, test harness updates

koivunej:
This adds [relevant](https://github.com/stepancheg/rust-protobuf/blob/master/src/lib/descriptor.rs#L4-L8) clippy and rustfmt ignores to generated output. There is one problem though:

```
$ cargo test
   Compiling quick-protobuf v0.4.0 (file:///home/joonas/src/quick-protobuf)
error: expected item after attributes
 --> tests/rust_protobuf/v2/test_group_pb.rs:8:34
  |
8 | #[cfg_attr(rustfmt, rustfmt_skip)]
  |                                  ^
error: aborting due to previous error
```

It errors out because `tests/rust_protobuf/v2/test_group_pb.proto` does not produce any messages as in the generated file contains only attributes, and apparently `cfg_attr` cannot be present without any rust code. Could this be a missed test case or something?


tafia:
Well, groups are not supported so indeed there is no rust code (and no tests).
We should probably check if there is something to write first.

koivunej:
> I don't plan to add group support anytime soon because it has been deprecated and I prefer spending time elsewhere. but if you (or anyone reading this) want it I welcome any PR.

I have no need for this. I guess I could just remove the problematic file then, perhaps add an error for unsupported feature?

Now that I'm back at my original project, looks like we probably want to add `#![allow(dead_code)]`, otherwise every messages `from_reader` will trigger a warning.

koivunej:
My latest push handles that problematic codegen file as "should fail". In the CI report there should now be the following files as failing the codegen:

 * `../tests/rust_protobuf/v2/test_root_pb.proto`
 * `../tests/rust_protobuf/v2/test_group_pb.proto` (this should had been handled but apparently not)
 * `../tests/rust_protobuf/v3/test_enum_alias_pb.proto`
 * `../tests/rust_protobuf/v2/test_expose_oneof_pb.proto`
 * `../tests/rust_protobuf/v2/test_enum_alias_pb.proto`

Can you comment on the rest of the files, should they codegen ok?

tafia:
This is very nice thanks.
I well less versed in shell scripts than you 😄 

I have 2 remarks:
- we could extend the list of `must fail` 
  - ../tests/rust_protobuf/v2/test_root_pb.proto : root search is not implemented yet
  - ../tests/rust_protobuf/v3/test_enum_alias_pb.proto: enum alias not implemented
  - ../tests/rust_protobuf/v2/test_enum_alias_pb.proto: same
  - ../tests/rust_protobuf/v2/test_expose_oneof_pb.proto: this one is a known issue (there should be some missing file in rust-protobuf repo, I didn't spend time figuring this out yet)
- For the issue with `#[cfg_attr(rustfmt, rustfmt_skip)]` on empty file: we should probably delete the generated file if there is an error ... or rename it to something like `file.rs.broken` ?

tests/rust_protobuf/v2/test_enum_values_pb.rs
```diff
@@ -3,6 +3,9 @@
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
+#![allow(unknown_lints)]
+#![allow(clippy)]
+#[cfg_attr(rustfmt, rustfmt_skip)]
```
overvenus: Why not `#![cfg_attr(rustfmt, rustfmt_skip)]`? It applies to the item enclosing it, in this case the item is `mod test_enum_values_pb`. Also I guess it will fix `error: expected item after attributes`.

See more: https://doc.rust-lang.org/book/attributes.html

koivunej: Phew, what a typo.

koivunej:
With the latest push I've:
 * added comments around the bash script
 * added the failures you described
 * removed the non-generating `.rs` files and unlisted the modules
 * fixed the `cfg_attr` typo (!!!)

As the CI runs now complete I think WIP can be removed and this thing is ready to be merged.

> I am well less versed in shell scripts than you :smile:

Not sure if it's a good thing :) Perhaps one day someone comes along and replaces the shell script with more readable rust code but I think the beefed up version will do for some time. One critical issue I fixed in the shell scripts is that they ignored all failures that did not happen on the last executed command (as sh/bash does without `set -e`).

tafia:
Merging!

codegen/src/errors.rs
```diff
@@ -22,5 +22,8 @@ error_chain! {
             display("cannot convert protobuf import into module import: {}", desc)
             cause("import definition might be invalid, some characters may not be supported")
         }
+        EmptyRead {
+            description("No messages or enums were read; either there was no input or there were only unsupported structures")
```
koivunej: Sorry I did not see this comment!

I'm not exactly sure why you can write `cause(...)` with `error_chain` but I understood that `cause`s are only for wrapped errors. You could for example have a call a function that returns `std::io::Error` and use `ResultExt::chain_err` to wrap it into more understandable, only for printing purposes though. `error_chain` does not support introspection with chained errors using `match`. I think line 23 works because `&'static str` is acceptable as an inner `Error + 'static + Send + Sync` using some magic but I do not think this is the right way to use it as a sort of "more details". Though, I cannot imagine why it would be there in the macro if it wasn't for this.

My understanding comes from [ResultExt::chain_err](https://docs.rs/error-chain/0.8.1/error_chain/example_generated/trait.ResultExt.html#tymethod.chain_err), [ErrorChainIter](https://docs.rs/error-chain/0.8.1/error_chain/struct.ErrorChainIter.html) and the original [Error::cause](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.cause) but I find the `error_chain` docs a bit hard to follow at some points.

Did some more digging on this and found that `error_chain` uses [`quick_error`](https://github.com/tailhook/quick-error) underneath (macro calling macro, calling macro, ...) and `quick_error`s README.md has an example where `cause(err)` is used to make `Error::cause` return the value inside the ErrorKind enum variant. So not meant for details.

Implementing the "more details" could probably be done with something like:

```
trait DetailedError {
    fn detailed_description(&self) -> Option<&str>;
}

impl DetailedError for errors::Error {
    fn detailed_description(&self) -> Option<&str> {
        match *self {
          (errors::ErrorKind::EmptyRead, _) => Some("Either there was no input or there were only unsupported structures"),
          (errors::ErrorKind::InvalidImport, _) => Some("..."),
          _ => None
    }
}

/// Newtype for formatting detailed errors, as we cannot override
/// their Display implementation generated by `error_chain`
struct DetailedDisplay<E: std::error::Error + DetailedError>(E);

impl std::fmt::Display for DetailedDisplay<E: std::error::Error + DetailedError> {
    fn fmt(...) -> std::fmt::Result {
        // like normal display for Errors, but with the possible details
    }
}
```

But above is just one way this could be implemented. If this is really wanted, I guess first step could be to investigate how do `rustc` errors work as they have all kinds of hints and suggestions as details.

tafia:
Thanks for the details. Don't worry it's fine!

PR #56, opened 2017-02-24
document message vs length prefixed messages

### LINKED ISSUE #52: Docs: add example of decoding non-length prefixed message
This works with latest `rust-protobuf`, test case can be found in gist:

https://gist.github.com/koivunej/68321c230287f95f88816c278d721d16

It'd seem that problem is in `reader::BytesReader::read_message` assumes length delimited contents. My use case is to read protobuf messages from inside an already framed stream. I can't seem to find any support for reading without the length prefix, so I guess that needs to be added?

tafia:
Closes #52 

@koivunej let me know if it looks ok. I know it's been a long time but in case you have some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,18 @@ fn main() {
     let mut reader = Reader::from_file("/path/to/binary/protobuf.bin")
         .expect("Cannot read input file");
     
-    // use the generated module fns with the reader to convert your data into rust structs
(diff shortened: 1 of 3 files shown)


koivunej:
Just read through it and looks good to me. Good idea to put this into readme as well. Have to admit that the readme was already extensive enough for my use cases that I haven't browsed the docs for this crate so much..

PR #57, opened 2017-02-27
pb-rs reworking

stevedonovan:
The `INCLUDE_PATH` feature is currently not quite working, but I thought that I'd show some progress for review in the meantime. I've tried to make all errors pass through error-chain (i.e. any panics should be considered as bugs). Multiple .proto files can be specified, and a destination directory specified.

The main change is that all modules use the Rust conventions. For example, if some file has a package `a.b` then (1) it will be called `b.rs` and (2)  `a/mod.rs` will be created/updated accordingly. The only explicit use of `mod` is for the scopes of nested items. (This required a somewhat clumsy solution, not yet happy.)

Tests pass, including my more elaborate examples.


stevedonovan:
Cool, I'll look at these issues and restore test sanity.

Some of my more elaborate examples had hex literals (`0x1F`) for enum values, so I'll try to teach the parser about them - it is part of the expected spec.  (I work with embedded programmers - they think in hex)

codegen/src/types.rs
```diff
@@ -749,24 +751,24 @@ impl Message {
         Ok(())
     }
 
-    fn set_package(&mut self, package: &str) {
+    fn set_package(&mut self, package: &str, module: &str) {
```
stevedonovan: It's a hack, and so needs review. Since `mod_` is not prefixed to module names, I needed a way of ensuring that _nested_ messages did actually sit in a `mod_MESSAGE` module. Hence the separate package (`a.b.c.C.D`) vs module (`a.b.c.mod_C.D`).

codegen/src/types.rs
```diff
@@ -749,24 +751,24 @@ impl Message {
...
         // set package = current_package.package.name to nested messages
-        if package.is_empty() {
-            for m in &mut self.messages { m.set_package(&self.name); }
-            for m in &mut self.enums { m.set_package(&self.name); }
-            for m in &mut self.oneofs { m.set_package(&self.name); }
+        let sub = format!("mod_{}",self.name);
+        let (prefix,module) = if package.is_empty() {
```
stevedonovan: You're right. It's really the one change I feel unhappy about.

codegen/src/types.rs
```diff
@@ -749,24 +751,24 @@ impl Message {
...
+            (self.name.clone(),sub)
         } else {
             self.package = package.to_string();
-            let child_package = format!("{}.{}", package, self.name);
-            for m in &mut self.messages { m.set_package(&child_package); }
-            for m in &mut self.enums { m.set_package(&child_package); }
-            for m in &mut self.oneofs { m.set_package(&child_package); }
-        }
+            self.module = module.to_string();
+            (format!("{}.{}", package, self.name),format!("{}.{}", module, sub))
```
stevedonovan: Yeah, that looks better!

stevedonovan:
PR #2 should be much better!

tafia:
I guess you're speaking about a new PR you'll be doing because i don't see the link between this PR and #2.
Thanks for your work

PR #59, opened 2017-03-14
pb-rs second revision

stevedonovan:
Apart from changes to codegen, I've added hex literals to enums and allowed semi-colons after enums & messages in the parser.

There's also a `-n` option for just dumping the names of enums and messages found in a file.

The challenge for this revision was coping with protobuf definitions with explicit packages and those without. In both cases, the generated .rs file implicitly defines a module, so no-package protos get an implicit module. So the `package` and `module` fields now differ in two ways

  -  a package like `a.A` (where `A` is a message scope) becomes module `a.mod_A`
  - no package in a file called `one.proto` has a module `one`

As before, the idea is that the generated .rs files must match the (sanitized) module names. So a number of test proto files now generate different Rust files. All the tests now pass, with that proviso.


tafia:
Awesome!
I'll have a look asap

stevedonovan:
I'm chuffed - tests just passed!

tests/rust_protobuf/v2/test_sync.rs
```diff
@@ -1,35 +0,0 @@
-use std::thread;
```
tafia: file deleted

tests/rust_protobuf/v2/test_text_format.rs
```diff
@@ -1,95 +0,0 @@
-use std::borrow::Cow;
```
tafia: file deleted

tests/rust_protobuf/v3/mod.rs
```diff
@@ -1,15 +1,4 @@
-mod test_basic_pb;
-mod test_basic;
-
-// mod test_zeros_are_not_written;
```
tafia: similarly that before, we should keep track of missing features somewhere

tests/rust_protobuf/v3/test_basic.rs
```diff
@@ -1,240 +0,0 @@
-use quick_protobuf::*;
```
tafia: file deleted

tests/rust_protobuf/v3/test_map.rs
```diff
@@ -1,48 +0,0 @@
-use quick_protobuf::*;
```
tafia: file deleted

tests/rust_protobuf/v3/test_zeros_are_not_written.rs
```diff
@@ -1,13 +0,0 @@
-use quick_protobuf::*;
```
tafia: file deleted

codegen/src/main.rs
```diff
@@ -26,41 +28,98 @@ fn main() {
...
+        let mut out_file = in_file.with_extension("rs");
+
+         if let Some(ofile) = matches.value_of("OUTPUT") {
+            out_file = PathBuf::from(ofile);
+        }
 
-    let out_file: PathBuf = matches.value_of("OUTPUT")
-        .map(|a| a.to_string().into())
-        .unwrap_or(in_file.with_extension("rs"));
+        if let Some(dir) = matches.value_of("OUTPUT_DIR") {
```
stevedonovan: Yes, it would make more sense - they're mutually exclusive possibilities.__

codegen/src/main.rs
```diff
@@ -26,41 +28,98 @@ fn main() {
...
+fn extension_matches<P: AsRef<Path>>(path: P, expected: &str) -> std::result::Result<(), String> {
     match path.as_ref().extension() {
         Some(x) if x == expected => Ok(()),
         Some(x) => Err(format!("Expected path with extension '{}', not: '{}'", expected, x.to_string_lossy())),
         None => Err(format!("Expected path with extension '{}'", expected)),
     }
 }
+
+fn path_vec(maybe_vec: std::result::Result<Vec<String>,clap::Error>) -> Vec<PathBuf> {
+    maybe_vec.unwrap_or(Vec::new()).iter().map(|s| s.into()).collect()
```
stevedonovan: Sure. I thought it would be premature optimization to worry about a few vectors at this point ;)

codegen/src/parser.rs
```diff
@@ -164,27 +166,41 @@ named!(message<Message>,
...
+        ) >> (num)
+    ));
+
+named!(integer<i32>,  map_res!( map_res!(digit,str::from_utf8), str::FromStr::from_str));
+
+named!(enum_field<(String, i32)>,
        do_parse!(name: word >> many0!(br) >>
                  tag!("=") >> many0!(br) >>
-                 number: map_res!(map_res!(digit, str::from_utf8), str::FromStr::from_str) >> many0!(br) >>
+                 number: alt!(hex_integer | integer) >> many0!(br) >>
```
stevedonovan: The people I work with are EEs with C++ background. So they speak hex as their mother tongue, and tend to throw in random semi-colons!

codegen/src/types.rs
```diff
@@ -540,12 +541,13 @@ pub struct Message {
     pub package: String,        // package from imports + nested items
     pub messages: Vec<Message>, // nested messages
     pub enums: Vec<Enumerator>, // nested enums
+    pub module: String,
```
stevedonovan: Cool

codegen/src/types.rs
```diff
@@ -1064,91 +1081,137 @@ impl FileDescriptor {
...
+            for path in import_search_path {
+                let candidate = if path.is_absolute() {
+                    path.join(&import)
+                } else {
+                    in_file.parent()
+                        .map_or_else(|| path.join(&import),
+                         |p| p.join(path).join(&import))
+                };
+                if candidate.exists() {
+                    matching_file = Some(candidate);
```
stevedonovan: We definitely should!

examples/codegen/data_types.rs
```diff
@@ -77,7 +78,7 @@ pub struct FooMessage<'a> {
     pub f_sint32: Option<i32>,
     pub f_sint64: i64,
     pub f_bool: bool,
-    pub f_FooEnum: Option<FooEnum>,
+    pub f_FooEnum: Option<data_types::FooEnum>,
```
stevedonovan: And this is EXACTLY why such files needed a `use self::*` so they could refer to items within the current module. Very hacky! I considered this and found it surprisingly tricky, although worth revisiting. The problem is that at the point of expansion, we don't know the package of the message containing the field.

tests/rust_protobuf/v2/test_basic.rs
```diff
@@ -1,240 +0,0 @@
-use quick_protobuf::*;
```
stevedonovan: ouch!

codegen/src/main.rs
tafia: I agree, this is just that the PR was so big I'm not sure I will review it in details many times. So just not to forget. It is really not important

examples/codegen/data_types.rs
tafia: Hmm ok
Let's merge it like it first and see later if we can work around this

stevedonovan: Let's hang for a few days to get this right. I've started doing a refactoring where I pass a reference to `FileDescriptor` everywhere (rather than separately passing messages and enums) and yes, then enough information is available. It also expresses the dependency on the enclosing file descriptor explicitly.  And let me sort out the missing files so everything merges cleanly!

tafia:
Thanks for all the time you've spent on this!

stevedonovan:
It's cool - a very interesting project that I find useful and refactoring is entertaining!  There are a few unstable features which I'm watching which would help here - like when you have to iterate over all message fields plus the oneof fields.  `-> impl Iterator` will make that much cleaner by allowing it to be pulled out as a method. I think we agree that macros are ugly for that kind of thing!

tafia:
the `Impl trait` is definitely something that I'm waiting for to enhance several libraries of mine!

tafia:
@stevedonovan I've just fixed the issue with appveyor.
Can you rebase?

tafia:
What is the status for this PR? Do you want me to jump in?

stevedonovan:
I'll finish the PR off and do the rebasing - give me a day or two!

stevedonovan:
So how does that work for you now? It is not the most _pretty_ code, but it passes the tests ;)  We can always make it pretty later. (My pride will probably make that happen.)

tafia:
This looks great!

But I'd like all tests to be put back to guarantee we do not have a regression.
There are several files have been removed/replaced (here are some example):
- tests/rust_protobuf/v2/test_basics.rs
- tests/rust_protobuf/v2/test_default_values_pb.rs (and this one has been overriden:  tests/rust_protobuf/v2/test_default_values.rs )
- tests/rust_protobuf/v2/test_enum_values.rs
- tests/rust_protobuf/v2/test_lite_runtime_pb.rs 
- tests/rust_protobuf/v3/test_zeros_are_not_written.rs 

Sometimes it is because the test file name conflicts with the generated name, in that case you can rename the original test file.

tafia:
Hi @stevedonovan just wondering if you plan on finishing this PR (ie put back all the tests) or you prefer handling it to someone else. I understand it's been a long time and you probably have other urgent topics right now!

stevedonovan:
Sorry, Johann! Will definitely sort things out this weekend - got a little
too busy.

tafia:
No worries 😄

stevedonovan:
Making progress, but then people decided to have a barbeque in winter. So another day or so I should be up-to-date with the review, I can see all the points now.

stevedonovan:
Hi Johann, I've resolved the test issues, but Github is confusing me here. I've updated my master.

tafia:
Thanks!
I'll have a look asap

stevedonovan:
The test issue is that whenever a .proto file defines an _explicit_ module, that module is used as the generated Rust file name, since this revision uses the Rust module system for scoping. In the process I got rid of a lot of quite legitimate test files!  Also had to edit the imports in the test files - generally they are simpler!

codegen/src/types.rs
```diff
@@ -1160,10 +1241,12 @@ impl FileDescriptor {
             }
         }
         // this is very inefficient but we don't care ...
-        let msgs = self.messages.clone();
+        //let msgs = self.messages.clone();
```
tafia: could you remove these commented lines?

codegen/src/types.rs
```diff
@@ -1177,10 +1260,11 @@ impl FileDescriptor {
 
     fn set_enums(&mut self) {
         // this is very inefficient but we don't care ...
-        let msgs = self.messages.clone();
+        //let msgs = self.messages.clone();
```
tafia: to remove

codegen/src/types.rs
```diff
@@ -1203,7 +1267,7 @@ impl FileDescriptor {
         writeln!(w, "#![allow(non_snake_case)]")?;
         writeln!(w, "#![allow(non_upper_case_globals)]")?;
         writeln!(w, "#![allow(non_camel_case_types)]")?;
-
+        writeln!(w, "#![allow(unused_imports)]")?;
```
tafia: I believe we will open an issue later to find out if we can remove some of these attributes automatically in the future. Let's not block the PR

codegen/src/types.rs
```diff
@@ -1237,51 +1298,28 @@ impl FileDescriptor {
...
+        }
+        write!(w, "use ")?;
+        for _ in 0..depth {
+            write!(w,"super::")?;
+        }
+        writeln!(w,"*;")?;
         Ok(())
     }
 
     fn write_package_start<W: Write>(&self, w: &mut W) -> Result<()> {
```
tafia: still a valid question?

codegen/src/types.rs
```diff
@@ -1356,3 +1410,64 @@ fn break_cycles(messages: &mut [Message], leaf_messages: &mut Vec<String>) {
...
+    let matches = "pub mod ";
+    let mut present = false;
+    let mut exists = false;
+    if let Ok(f) = File::open(&file) {
+        exists = true;
+        for line in BufReader::new(f).lines() {
+            let line = line?;
+            if let Some(i) = line.find(matches) {
+                let rest = &line[i+matches.len()..line.len()-1];
+                if rest == name {
```
tafia: what if you have `pub use foobar;` and you are looking for `pub use foo;`?

codegen/src/types.rs
```diff
@@ -1356,3 +1390,64 @@ fn break_cycles(messages: &mut [Message], leaf_messages: &mut Vec<String>) {
...
+        if let Some(i) = package.rfind('.') {
+            (&package[0..i],&package[i+1..])
+        } else {
+            ("",package)
+        }
+    }
+}
+
+/// Given a file path, create or update the mod.rs file within its folder
+fn update_mod_file(path: &Path) -> Result<()> {
```
tafia: This is still a concern for me:
- either the mod.rs has been built by the codegen: in this case **we know up-front all the necessary `pub mod` we need**
- or the mod.rs is an existing file: then we are potentially doing things wrong here!:
  - not only we are modifying existing files (dangerous!)
  - we are appending the new uses at the end of the file (ie after the code ...)

tafia:
@stevedonovan have you seen my comments? We are really close to have this merged and the generated files will look much better, anything I can do to help?

stevedonovan:
I've seen the comments, thanks! Hope to work on them later this evening, so we can close this.

codegen/src/parser.rs
```diff
@@ -139,12 +141,13 @@ named!(message_event<MessageEvent>, alt!(reserved_nums => { |r| MessageEvent::Re
...
-named!(message_events<(String, Vec<MessageEvent>)>, 
-       do_parse!(tag!("message") >> many0!(br) >> 
-                 name: word >> many0!(br) >> 
+named!(message_events<(String, Vec<MessageEvent>)>,
+       do_parse!(tag!("message") >> many0!(br) >>
+                 name: word >> many0!(br) >>
                  tag!("{") >> many0!(br) >>
                  events: many0!(message_event) >>
                  many0!(br) >> tag!("}") >>
+                 many0!(br) >> many0!(tag!(";")) >>
```
stevedonovan: Like I say, people tend to throw in random semi-colons ;)

codegen/src/types.rs
```diff
@@ -170,7 +170,7 @@ impl FieldType {
 
                 if found.is_none() {
                     // recursively search into nested messages
-                    for m in msgs {
+                    for m in msgs.iter() {
```
stevedonovan: Ah, I got into a habit of being explicit about iterators. But you're quite correct, and I'd like to follow your style.

codegen/src/types.rs
```diff
@@ -209,22 +209,22 @@ impl FieldType {
         }
     }
 
-    fn has_lifetime(&self, msgs: &[Message], packed: bool) -> bool {
+    fn has_lifetime(&self, desc: &FileDescriptor, packed: bool) -> bool {
```
stevedonovan: Totally for that reason only!

codegen/src/types.rs
```diff
@@ -1064,91 +1098,138 @@ impl FileDescriptor {
...
-            }).collect();
+        let (prefix,file_package) = split_package(&desc.package);
+        if ! file_package.is_empty() {
+            file_stem = file_package.to_string();
             sanitize_keyword(&mut file_stem);
-            config.out_file.as_ref().with_file_name(format!("{}.rs", file_stem))
-        };
-        let mut w = BufWriter::new(File::create(out_file)?);
-        desc.write(&mut w, name)
+        }
```
stevedonovan: Nice point - it still needs to be `mut` because of the borrow...

codegen/src/types.rs
```diff
@@ -1064,91 +1098,138 @@ impl FileDescriptor {
...
+
+        if ! prefix.is_empty() {
+            use std::fs::create_dir_all;
+            // e.g. package is a.b; we need to create directory 'a' and insert it into the path
+            let file = PathBuf::from(out_file.file_name().unwrap());
+            out_file.pop();
+            for p in prefix.split('.') {
+                out_file.push(p);
+            }
+            if ! out_file.is_dir() {
```
stevedonovan: Well caught - `exists` is what I mean.

codegen/src/types.rs
stevedonovan: The slice for `pub use foobar;` is going to be "foobar", which is matched exactly. This feels less than elegant, but it works as intended.  (that `line.len()-1` feels very hacky)

stevedonovan:
Right, I'll finish up tomorrow with the `mod.rs` question. I agree that we should not overwrite files which were _not_ written by us.  Because of the design (originally designed for processing a file at a time) we don't know the full structure of the generated code - and it's always possible to add new proto files. So I propose that we mark our `mod.rs` files very distinctly with a `// GENERATED BY PB-RS' comment up front - if we find that, then we can modify away.

stevedonovan:
Hi Johann!  I've taken the simplest possible route to a solution - only modify `mod.rs` if it is one of ours. They are generated with a first line `//! Automatically generated mod.rs` and we only touch such files! In fact, I caught a case where it _was_ overwriting one of the test files, so you were right. Parts are not so very elegant, but I guess we go for 'correct', 'robust' and then 'elegant' (although maybe 'gives better parsing errors' should be more of a priority now. I can look at the Nom parser).  

That `use super::*` is still a little irritating, since I cannot yet only write it when it will be used, but we can always use a '#![...]' directive for this offending line only.

tafia:
I finally merged this!
Thanks again for all the energy you put in this! This is a very valuable improvement.
I'll open an issue about the header trick and to see if you can get rid of these `use super::*`

stevedonovan:
This is very cool! Sorry, I was busy being ill with flu.  I think that the main thing is that code that _uses_ generated modules should experience as little friction as possible, but a certain amount of ugliness is forgivable in generated code. But let me have a look!

PR #70, opened 2017-09-07
Correct unnecessary boxing of enums

### LINKED ISSUE #69: unnecessary boxing for enums
Consider:
```
enum City {
    LONDON = 0;
    PARIS = 1;
}

message Address {
    City city = 1;
}

message Person {
    Address address = 2;
}
```

Generates:
```
pub struct Person {
    pub address: Option<Box<Address>>,
}
```

Expected code:
```
pub struct Person {
    pub address: Option<Address>,
}
```

Also consider:
```
message Address {
    enum City {
        LONDON = 0;
        PARIS = 1;
    }
    required City city = 1;
}

message Person {
    oneof specification {
        Address address = 1;
    }
}
```

Generates:
```
pub struct Person {
    pub specification: mod_Person::OneOfspecification,
}

impl Person {
    pub fn from_reader(r: &mut BytesReader, bytes: &[u8]) -> Result<Self> {
        let mut msg = Self::default();
        while !r.is_eof() {
            match r.next_tag(bytes) {
                Ok(10) => msg.specification = mod_Person::OneOfspecification::address(Box::new(r.read_message(bytes, Address::from_reader)?)),
                Ok(t) => { r.read_unknown(bytes, t)?; }
                Err(e) => return Err(e),
            }
        }
        Ok(msg)
    }
}
```

Which doesn't compile. The specification of Person doesn't contain a box, but on the line starting with `Ok(10) ` it attempts to box something.

sDessens:
The code generator attempts to find cycles in the protobuf specification
by recursively marking nodes that are not part of a cycle as leaf nodes,
and then inserting Box<> here-and-there until no more cycles exists. But
the function forgets to consider enums as leaf nodes, causing
unnecessary boxing and in some cases broken code.

Fixes #69

### DIFF
--- a/codegen/src/types.rs
+++ b/codegen/src/types.rs
@@ -1095,7 +1095,7 @@ impl FileDescriptor {
             bail!(ErrorKind::EmptyRead);
         }
 
-        let mut leaf_messages = Vec::new();
+        let mut leaf_messages = desc.enums.iter().map(|e| e.name.to_string()).collect::<Vec<_>>();
         break_cycles(&mut desc.messages, &mut leaf_messages);
 
         desc.sanity_checks()?;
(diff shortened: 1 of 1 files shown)


tafia:
closing for #71

PR #71, opened 2017-09-11
Set enums before break_cycles

### LINKED ISSUE #69: unnecessary boxing for enums

tafia:
Closes #69 

@sDessens do you confirm it works as expected?

cc @stevedonovan @koivunej

### DIFF
--- a/codegen/src/types.rs
+++ b/codegen/src/types.rs
@@ -1086,7 +1086,6 @@ pub struct FileDescriptor {
 
 impl FileDescriptor {
 
-
     pub fn write_proto(config: &Config) -> Result<()> {
         let mut desc = FileDescriptor::read_proto(&config.in_file, &config.import_search_path)?;
 
--- a/codegen/src/types.rs
+++ b/codegen/src/types.rs
@@ -797,7 +797,7 @@ impl Message {
 
     /// Searches for a matching message in all message
     ///
-    /// If none is found,
+    /// If none is found, then it is an enum
     fn set_enums(&mut self, desc: &FileDescriptor) {
         for f in self.fields.iter_mut()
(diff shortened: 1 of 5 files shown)


koivunej:
It has been a while since I generated or did anything with eventstore-tcp (pending a tokio 0.2 release) and I was never hit by anything like this but fix seems good with a testcase so :+1:.

tafia:
thanks

PR #76, opened 2017-09-26
corrected find_enum and find_message for keyword identifiers

awestlake87:


tafia:
First thanks for the PR and sorry for answering this late.

Can you please give more context why it is needed. Could you add a broken test for instance?
As for your fn here is a *rustier* version of if, (could be enhanced but performance of codegen is really not critical.

```rust
pub fn get_ident(ident: &str) -> String { // `&String` is usually replaced by `&str` which both point to the same data
    if !ident.contains('.') {
        if RUST_KEYWORDS.contains(&&ident[..]) {
            format!("{}_pb", ident) // this will create a new `String`, probably without new allocation
        }
        else {
            ident.clone()
        }
    }
    else {
        ident.split('.').map(|s| get_ident(s)).collect::<Vec<_>>().join(".") // not pretty but ok!, note that i could remove the `to_string` because I changed the signature to work with `&str`
    }
}
```

PR #82, opened 2017-11-15
Replace from_reader impl with MessageRead trait

### LINKED ISSUE #78: Possibility of a MessageRead trait?
This isn't a big problem, but I was wondering if there's any technical reason why from_reader is generated as an independent method rather than on some MessageRead trait.

If it were a trait, it would be possible to simplify reading code - and not have to specify `TypeName::from_reader` as an argument to `BytesReader.read_message`. It would also allow consumers to write generic utility methods useful for all of their messages.

Is there a specific advantage to the way quick-protobuf does this now that I'm not seeing?

daboross:
This has some advantages with clarity in the type system, but the disadvantage of disallowing custom 'read' functions passed to `BytesReader::read_message`.

This also adds a need for a turbofish in some cases of `reader.read_mesage::<TypeName>(...)`.

Really liking the library so far, and hoping this is an overall positive change!

Closes #78.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,15 +82,17 @@ fn main() {
     // Use the generated module fns with the reader to convert your data into rust structs.
     //
     // Depending on your input file, the message can or not be prefixed with the encoded length
-    // for instance, a *stream* which contains several messages generally split them using this 
+    // for instance, a *stream* which contains several messages generally split them using this
     // technique (see https://developers.google.com/protocol-buffers/docs/techniques#streaming)
     //
     // To read a message without a length prefix you can directly call `FooBar::from_reader`:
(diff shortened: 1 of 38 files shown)


tafia:
Awesome!
I'll have a look asap.

daboross:
Thanks!

Just realized I had a documentation error on MessageRead, just pushed a fix for that.

tafia:
I close this. I have rebased it in #85. Once travis is happy i'll merge it.
Thanks again!

daboross:
Thank you! Will make it nicer to use I hope.

tafia:
yeah!
Just published v0.6.0.

PR #98, opened 2018-04-22
Provide From<&str> impls for enums

edef1c:


edef1c:
I'm unsure about the merits of this versus an impl of `FromStr<Error=!>`, either seems reasonable.

tafia:
Thanks!
`From` is fine

tafia:
Published pb-rs 0.3.1

PR #102, opened 2018-07-11
Explicitly mention the usage of pb-rs as a crate

Moxinilian:
When choosing what crate to use for protobuf support, I hesitated to use quick-protobuf as I believed pb-rs was only usable through the command line. When I discovered that it is actually usable as a crate, I had already started to use another crate, that isn't as good as quick-protobuf.

To avoid this to happen again, this PR explicitly mentions in the README.md for pb-rs that it is usable as a crate, alongside a simple example.

### DIFF
--- a/codegen/README.md
+++ b/codegen/README.md
@@ -7,3 +7,24 @@ A simple converter from .proto files into rust quick-protobuf compatible modules
 ```
 pb-rs <file.proto>
 ```
+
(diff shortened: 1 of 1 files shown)


Moxinilian:
I think CI is failing in nightly for something I am not responsible of.

tafia:
Thanks! I'll fix the travis tests asap.

PR #105, opened 2018-07-12
Allow to advance arbitrary bytes in BytesReader

Moxinilian:
I am working on a network protocol using protobuf, and the buffers I get are protocol buffers with a 2 bytes packet ID header. However, there seems to be no good way to prevent the reader from taking those two bytes into the decoding process.

This PR provides a way to skip an arbitrary amount of bytes when using BytesReader, so that messages with additional data in the same buffer can be read more easily.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -7,15 +7,15 @@
 //!
 //! It is advised, for convenience to directly work with a `Reader`.
 
+use std::fs::File;
 use std::io::{self, Read};
 use std::path::Path;
-use std::fs::File;
 
 use errors::{Error, Result};
 use message::MessageRead;
 
-use byteorder::LittleEndian as LE;
 use byteorder::ByteOrder;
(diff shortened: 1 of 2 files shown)


tafia:
Thanks for the PR!

I'd like having some explanation in the comment because this is not something we do in general and it might confuse people. Could you add an example where this is useful?

Also I'll probably merge #104, there may be some conflict with your code, if this is the case could you rebase?

Moxinilian:
Apparently there seem to be no conflicts.
Maybe renaming the method to `skip` would help?
Unfortunately I don't see a context-free example for this. It works well in the context of skipping header data for example (unless you find a better alternative for my specific case, and so I would gladly close this PR and implement it instead).

Moxinilian:
I think making `read_byte` public might solve all issues where skipping is needed.
I think it would be a better approach. Should I close this PR or is this method still useful to have?

PR #106, opened 2018-07-13
Make byte-specific utilities available in BytesReader

Moxinilian:
See #105 for motivations.

This PR makes `read_u8` public in the BytesReader, and creates a `write_u8` method in the writer.
This will allow to write multiple messages and separate them in a single buffer with a single writer.


tafia:
Thanks!

tafia:
Published v0.6.1

Issue #97, opened 2018-04-21
Wrong field type info for a field with enum as default 


jackalcooper:
The `typ` of a field with a Enum default in a message is `FieldType::Message(String)` not `FieldType::Enum(String)`

Issue #100, opened 2018-04-30
Wrong FieldType for a Enum field 


jackalcooper:
The typ of a field with a Enum default in a message is FieldType::Message(String) not FieldType::Enum(String). The repeated field has this problem as well.

Issue #96, opened 2018-04-21
A enum field as default won't be checked


jackalcooper:
to_be_import.proto
```proto
enum Enum1 {
  a = 0;
  b = 1;
  c = 2;
}
```
```proto
import "to_be_import.proto"
message Msg1 {
  optional Enum1 x = 9 [default = a]; // a field exists
  optional Enum1 y = 9 [default = does_not_exist]; // a field doesn't exist
}
```
I also check google protoc out. It doesn't compile if this happens.

tafia:
Good point

PR #114, opened 2018-08-16
Make distinction between Message, Enum and MessageOrEnum

tafia:
MessageOrEnum being the state at the beginning of the parsing.
Also possibly fix some issue with enum in maps.

EDIT: also sanity checks that default enum values actually exist.


tafia:
closes #97 #100 #96

PR #119, opened 2018-08-28
don't check lifetime parameter on boxed fields

### LINKED ISSUE #118: Self-referential fields cause stack overflow
When I compile the following with `pb-rs`, I get a stack overflow:
```
/* -*- Mode: protobuf; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
 * vim: set ts=8 sts=4 et sw=4 tw=99: */

syntax = "proto2";

message ViaOneOf {
    oneof Alternatives {
        Data present = 1;
        bool absent  = 2;
    }

    message Data {
        required ViaOneOf self_reference = 1;
    }
}

message ViaOptional {
    optional Data data = 1;

    message Data {
        required ViaOptional self_reference = 1;
    }
}
```
Either definition alone is sufficient. The error is:
```
$ pb-rs Recursive.proto
Found 2 messages, and 0 enums
Writing message ViaOneOf

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted (core dumped)
$ 
```

tafia:
Close #118

### DIFF
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -659,6 +659,7 @@ impl Message {
         self.fields
             .iter()
             .chain(self.oneofs.iter().flat_map(|o| o.fields.iter()))
+            .filter(|f| !f.boxed) // removed boxed fields
             .any(|f| match f.typ {
                 FieldType::Message(ref m) if m.get_message(desc).name == self.name => false,
                 ref t => t.has_lifetime(desc, f.packed()),
@@ -1654,7 +1655,6 @@ impl FileDescriptor {
         self.write_enums(w)?;
         self.write_messages(w)?;
         self.write_package_end(w)?;
-        println!("Done processing {}", filename);
         Ok(())
     }
 
--- a/quick-protobuf/tests/rust_protobuf/v2/issue118.proto
+++ b/quick-protobuf/tests/rust_protobuf/v2/issue118.proto
@@ -1,0 +1,23 @@
+/* -*- Mode: protobuf; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+ * vim: set ts=8 sts=4 et sw=4 tw=99: */
+
+syntax = "proto2";
+
+message ViaOneOf {
+    oneof Alternatives {
+        Data present = 1;
+        bool absent  = 2;
+    }
+
+    message Data {
+        required ViaOneOf self_reference = 1;
+    }
+}
+
+message ViaOptional {
+    optional Data data = 1;
+
+    message Data {
+        required ViaOptional self_reference = 1;
+    }
+}


pb-rs/src/types.rs
```diff
@@ -1545,6 +1544,45 @@ impl FileDescriptor {
...
+            .messages
+            .iter()
+            .map(|m| m.index.clone())
+            .collect::<Vec<_>>();
+        while let Some(m) = messages.pop() {
+            messages.extend(m.get_message(self).messages.iter().map(|m| m.index.clone()));
+            processing.push((m.clone(), vec![m]));
+        }
+
+        while let Some((i, ancestors)) = processing.pop() {
```
jimblandy: This still fails on the following input:
```
/* -*- Mode: protobuf; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 4 -*-
 * vim: set ts=8 sts=4 et sw=4 tw=99: */

syntax = "proto2";

message First {
  required Third first_field = 1;
}

message Second {
  optional First second_field = 1;
}

message Third {
  required Second third_field = 1;
}
```

pb-rs/src/types.rs
```diff
@@ -659,6 +659,7 @@ impl Message {
         self.fields
             .iter()
             .chain(self.oneofs.iter().flat_map(|o| o.fields.iter()))
+            .filter(|f| !f.boxed) // removed boxed fields
```
jimblandy: This seems like a good approach.

tafia:
Thanks for the review. I have pushed another change which should work.
I have also rebased it.
Sorry for the buggy fix and thank you for your patience.

jimblandy:
Sure thing! I actually had a good time trying to find the counter-example...

pb-rs/src/types.rs
```diff
@@ -1540,34 +1547,41 @@ impl FileDescriptor {
...
+            debug!(
+                "break cycle:\nmessage: {:?}\nprocessing: {:?}",
+                i, processing
+            );
             for f in msg
-                .fields
-                .iter_mut()
-                .chain(msg.oneofs.iter_mut().flat_map(|o| o.fields.iter_mut()))
+                .all_fields_mut()
                 .filter(|f| !f.boxed && f.frequency != Frequency::Repeated)
```
jimblandy: Does the test for `!f.boxed` do anything here? The only place we ever set `.boxed` is in this function, which visits each message only once, and visits each of its fields only once. So it seems to me `.boxed` should always be `false` at this point.

jimblandy: So, I *think* this code is correct:

The `processing` queue contains exactly the set of not-yet-visited nodes. Visiting a node removes it from the queue forever, pulls its not-yet-visited direct successors to the front of the queue, and adds the path by which it was reached to their `ancestors` sets.

Because visiting a node pulls all its direct successors to the front of the queue, we visit nodes in a depth-first order: we always visit all not-yet-visited nodes reachable from that node before visiting any that are unreachable.

By the time the queue is empty, the entire graph is covered by one or more trees, in which each node's parent is the last of its predecessors to pull it to the front of the queue. Nodes never pulled to the front by any other are roots of their own trees.

Within one tree, each node's `ancestor` set contains all the nodes on the path to the root. Each node's parent is the node that pulled it to the front of the queue; and we always propagate the ancestors when we do so.

Every cycle in the graph has one node that was visited first, F. Since we visit every not-yet-visited node reachable from F before visiting any unreachable nodes, we will visit every node in the cycle before we visit any nodes unreachable from F. Hence, we will visit the node E that points to F. At this point, E's ancestor set will include F, and we will break the cycle.

jimblandy:
Although, I think there's a much simpler and more useful approach to this problem, based on the limitations of the protobufs language itself. I'll file a separate issue for that.

jimblandy:
Filed #121 to suggest alternative approach.

tafia:
Ok I have followed your advice on #121.
I have also written Tarjan's algorithm to detect all strongly connected component within which cycles are.
All optional fields are boxed within a cycle.

tafia:
Test are now failing because there are some cycles with required fields only. I personally think it is good that it fails (the proto file itself must be amended). Of course this is a big breaking change but it is because of unsound messages.

What do you think?

jimblandy:
Are these cycles of required fields only present in test input, or are they actually used in the real world? If it's just the tests, then it seems fine to fix the tests.

But if they're used in the real world, it's important for `pb-rs` to support existing usage. But, how is it even possible to send such a message? Are those message types all unused?

I noticed that `protoc` doesn't forbid cycles of required fields, so I guess I shouldn't be surprised that they exist in the wild. 

I haven't looked closely at the rest of `pb-rs`; does it ever print warnings? Perhaps the most helpful thing would be to always warn about cycles of required fields, to help people developing new protocols. Then, we could have options to silence the warnings, or treat them as errors, the way other compilers do.

PR #120, opened 2018-08-28
Abstract out iterating over a Message's Fields.

jimblandy:
Add methods `Message::all_fields` and `Message::all_fields_mut` that return
iterators over all a Message's fields, including those in `oneof` blocks. Use
these new methods to abstract out such iterations in the existing code.


tafia:
Thanks
Actually I was doing exactly the same thing yesterday along with some other changes.

jimblandy:
Okay, great!

PR #123, opened 2018-09-14
Allow build.rs to generate files in OUT_DIR

### LINKED ISSUE #122: It's hard for a build script to put pb-rs output in OUT_DIR
The `cargo package` command checks that the build script doesn't modify the source directory; as the Cargo book [explains][cb], build scripts are not supposed to modify anything outside the directory given in the `OUT_DIR` environment variable.

This rule is hard to follow when calling `pb-rs` from `build.rs`: the only way I'm aware of to incorporate code from OUT_DIR is via an `include!` directive, but that chokes on the `//!` comments that `pb-rs` places at the top of its generated files. When I write:

```
mod mozilla {
include!(concat!(env!("OUT_DIR"), "/mozilla/mod.rs"))
}
```
rustc complains:
```
error: macros that expand to items must either be surrounded with braces or followed by a semicolon
 --> src/dump/mod.rs:7:9
  |
7 | include!(concat!(env!("OUT_DIR"), "/mozilla/mod.rs"))
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[cb]: https://doc.rust-lang.org/cargo/reference/build-scripts.html

tafia:
close #122 

@jimblandy would it work for you?


jimblandy:
Yes, this would help, but I think it's more than I need!

I think the comments explaining what the Rust code was generated from could be quite valuable in large systems with more than a handful of generate files. The only problem that I know of is that they cannot be `//!` comments. So for my purposes, simply the one-character change to `write_headers` would suffice.

I'm not sure I would actually use the option to disable the headers, as long as they're just plain comments.

tafia:
It is not just with the comment, it is also with all the module attributes.

If the file is in the OUT_DIR folder, one should assume that it was automatically generated. Knowing that it is from pb-rs could be done at library level by just renaming the file for instance.

jimblandy:
Right, `//!` comments are just one special case of inner attributes in general. No inner attributes are allowed in a file you're going to `include!`. So it'd have to avoid generating not just doc comments, but any other sort of module attribute. That's too bad.

nerdrew:
Did you miss changing the `MAGIC_COMMENT` too?

tafia:
> Did you miss changing the MAGIC_COMMENT too?

Sorry I am not sure I follow. What is this?

nerdrew:
Sorry I wasn't more clear. The `MAGIC_COMENT` const still uses `//!`

PR #124, opened 2018-09-16
WIP add helper for using pb-rs from build.rs scripts

nerdrew:
Thoughts on adding a helper to make build.rs scripting easier?

### DIFF
--- a/pb-rs/src/lib.rs
+++ b/pb-rs/src/lib.rs
@@ -171,7 +171,6 @@ impl ConfigBuilder {
                     error_cycle: self.error_cycle,
                     headers: self.headers,
                 }
-            })
-            .collect()
+            }).collect()
     }
 }
--- a/pb-rs/src/lib.rs
+++ b/pb-rs/src/lib.rs
@@ -10,3 +10,125 @@ pub mod keywords;
 pub mod parser;
 pub mod scc;
 pub mod types;
+
+use errors::Error;
+use failure::ResultExt;
+use std::path::{Path, PathBuf};
+use types::{Config, FileDescriptor};
+
+#[derive(Debug, Default)]
+pub struct CompileBuilder {
+    in_files: Vec<PathBuf>,
+    out_file: Option<PathBuf>,
+    include_paths: Vec<PathBuf>,
+    single_module: bool,
+    no_output: bool,
+    error_cycle: bool,
+    headers: bool,
+}
+
+impl CompileBuilder {
+    pub fn new<P: AsRef<Path>>(
+        in_files: &[P],
+        output: Option<&P>,
+        output_dir: Option<&P>,
+        include_paths: &[P],
+    ) -> Result<CompileBuilder, ::failure::Error> {
+        let in_files = in_files
+            .iter()
+            .map(|f| f.as_ref().into())
+            .collect::<Vec<PathBuf>>();
+        let output = output.map(|f| f.as_ref().into());
+        let output_dir: Option<PathBuf> = output_dir.map(|f| f.as_ref().into());
+        let mut include_paths = include_paths
+            .iter()
+            .map(|f| f.as_ref().into())
+            .collect::<Vec<PathBuf>>();
+
+        if in_files.is_empty() {
+            return Err(Error::NoProto.into());
+        }
+
+        for f in &in_files {
+            if !f.exists() {
+                return Err(Error::InputFile(format!("{}", f.display())).into());
+            }
+        }
+
+        let out_file = match (output, output_dir) {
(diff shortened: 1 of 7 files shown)


pb-rs/src/lib.rs
```diff
@@ -10,3 +10,109 @@ pub mod keywords;
...
 pub mod scc;
 pub mod types;
+
+use errors::Error;
+use types::{Config, FileDescriptor};
+use std::path::PathBuf;
+use failure::ResultExt;
+
+#[derive(Debug, Default)]
+pub struct CompileBuilder {
```
tafia: I would prefer having a `ConfigBuilder` only. There is nothing *compiled* in my opinion.

pb-rs/src/lib.rs
```diff
@@ -10,3 +10,109 @@ pub mod keywords;
...
+    out_file: Option<PathBuf>,
+    include_paths: Vec<PathBuf>,
+    single_module: bool,
+    no_output: bool,
+    error_cycle: bool,
+    headers: bool,
+}
+
+impl CompileBuilder {
+    pub fn new(in_files: Vec<PathBuf>, output: Option<PathBuf>, output_dir: Option<PathBuf>, mut include_paths: Vec<PathBuf>) -> Result<CompileBuilder, ::failure::Error> {
```
tafia: Can you add comments for all these fns?

pb-rs/src/lib.rs
nerdrew: Sure. I'll rename it to `ConfigBuilder`.

nerdrew:
> I love the idea of using a builder. I would prefer limiting it to a config builder though ... and still have a FileDescriptor::run called.

I'm not sure what you mean by `FileDescriptor::run`. Are you thinking a new method that accepts a slice of `Config`s? Or perhaps the `ConfigBuilder` has an iter you can call to iterate over the generated configs and then call `FileDescriptor::write_proto`?

nerdrew:
> Could you also apply rustfmt (rustup component add rutfmt-preview) and add comments to all these builder fn (so they appear in the pb-rs docs).

LOL. I think my version of `rustfmt` might be different than yours. I had to disable it for this project since it changed every file when I saved :). Ya, I'll run it on the code I add.

nerdrew:
Something like?

```rust
impl ConfigBuilder {
    pub fn build(self) -> Vec<Config> {
        self.in_files
            .iter()
            .map(|in_file| {
                let mut out_file = in_file.with_extension("rs");

                if let Some(ref ofile) = self.out_file {
                    if ofile.is_dir() {
                        out_file = ofile.join(out_file.file_name().unwrap());
                    } else {
                        out_file = ofile.into();
                    }
                }

                Config {
                    in_file: in_file.to_owned(),
                    out_file: out_file,
                    import_search_path: self.include_paths.clone(),
                    single_module: self.single_module,
                    no_output: self.no_output,
                    error_cycle: self.error_cycle,
                    headers: self.headers,
                }
            }).collect()
    }
}
```

```rust
impl FileDescriptor {
    pub fn run(configs: &[Config]) -> Result<()> {
        for config in configs {
            Self::write_proto(&config)?
        }
        Ok(())
    }
}
```

```rust
FileDescriptor::run(&compiler.build()).map_err(|e| e.into())
```

tafia:
yes if you don't mind

nerdrew:
@tafia I updated this PR.

tafia:
Thanks!
Could you please rustfmt again?

nerdrew:
Ran: ```git diff --name-only upstream/master | grep -E '^pb-rs.*\.rs$' | xargs rustfmt```

pb-rs/src/types.rs
```diff
@@ -1879,7 +1890,7 @@ fn split_package(package: &str) -> (&str, &str) {
     }
 }
 
-const MAGIC_HEADER: &'static str = "//! Automatically generated mod.rs";
+const MAGIC_HEADER: &'static str = "// Automatically generated mod.rs";
```
nerdrew: This is unrelated to this PR, but was failing for me when using `include!`

tafia: Can you rebase? I think this was already done in a previous PR.

nerdrew:
My rustfmt:

```
% cargo +stable fmt --version
rustfmt 0.99.1-stable (da17b689 2018-08-04)
```

nerdrew:
I ran `cargo +stable fmt` from the root. I just re-ran it and nothing changed.

tafia:
Alright then, I am not sure what happened exactly.
Thanks!

PR #125, opened 2018-09-16
Correctly handle messages / enums that begin with .

josh33901:
As an example:

```protobuf
message CHTMLHeader {
	optional string key = 1;
	optional string value = 2;
}

message CHTMLPageSecurityInfo {
	optional bool bIsSecure = 1 [default = false];
	optional bool bHasCertError = 2 [default = false];
	optional string issuerName = 3;
	optional string certName = 4;
	optional int32 certExpiry = 5 [default = 0];
	optional int32 nCertBits = 6 [default = 0];
	optional bool bIsEVCert = 7 [default = false];
}

message CMsgFinishedRequest {
	optional uint32 browser_handle = 1;
	optional string url = 2;
	optional string pageTitle = 3;
	optional .CHTMLPageSecurityInfo security_info = 4;
	repeated .CHTMLHeader headers = 5;
}
```

`optional .CHTMLPageSecurityInfo` refers to global scope `message CHTMLPageSecurityInfo`
same with `CHTMLHeader`

Not sure if this is the most idiomatic way to do this in rust (I'm still relatively new)

### DIFF
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -1709,7 +1709,9 @@ impl FileDescriptor {
                     _ => vec![typ].into_iter(),
                 }) {
                 if let FieldType::MessageOrEnum(name) = typ.clone() {
-                    let test_names: Vec<String> = if m.package.is_empty() {
+                    let test_names: Vec<String> = if name.starts_with(".") {
+                        vec![name.clone().split_off(1)]
+                    } else if m.package.is_empty() {
                         vec![name.clone(), format!("{}.{}", m.name, name)]
                     } else {
                         vec![
--- a/quick-protobuf/tests/rust_protobuf/generate.sh
+++ b/quick-protobuf/tests/rust_protobuf/generate.sh
@@ -13,7 +13,6 @@ have_failures=""
 declare -A must_fail
 
 must_fail["v2/test_group_pb.proto"]="expected failure (empty read)"
-must_fail["v2/test_root_pb.proto"]="root search is not implemented yet"
 must_fail["v3/test_enum_alias_pb.proto"]="enum alias not implemented"
 must_fail["v2/test_enum_alias_pb.proto"]="enum alias not implemented"
 must_fail["v2/test_expose_oneof_pb.proto"]="missing file"


josh33901:
Unless i am missing something obvious why should `v2/test_root_pb.proto` not succeed?

tafia:
> Unless i am missing something obvious why should v2/test_root_pb.proto not succeed?

Just [not implemented yet :wink:](https://github.com/tafia/quick-protobuf/blob/master/quick-protobuf/tests/rust_protobuf/generate.sh#L16)

josh33901:
I'm not sure I entirely fixed it though as it might not work across imports. I'll check later.

josh33901:
> I'm not sure I entirely fixed it though as it might not work across imports. I'll check later.

Ok looks like it does work. I was getting confused between some messages that use messages dont compile at the moment.

tafia:
> I was getting confused between some messages that use messages dont compile at the moment.

Can you clarify?

josh33901:
> > I was getting confused between some messages that use messages dont compile at the moment.
> 
> Can you clarify?

Messages with extensions

```protobuf
import "google/protobuf/descriptor.proto";

option cc_generic_services = false;

extend .google.protobuf.EnumValueOptions {
	optional string network_connection_token = 50500;
}

enum ENetworkDisconnectionReason {
	NETWORK_DISCONNECT_INVALID = 0;
	NETWORK_DISCONNECT_SHUTDOWN = 1;
	NETWORK_DISCONNECT_DISCONNECT_BY_USER = 2 [(network_connection_token) = "#GameUI_Disconnect_User"];
	NETWORK_DISCONNECT_DISCONNECT_BY_SERVER = 3 [(network_connection_token) = "#GameUI_Disconnect_Server"];
	NETWORK_DISCONNECT_LOST = 4 [(network_connection_token) = "#GameUI_Disconnect_ConnectionLost"];
 // ...
}
```

in one file that is imported into another where it is used as
```protobuf
optional .ENetworkDisconnectionReason disconnect_reason = 2 [default = NETWORK_DISCONNECT_INVALID];
```

Which shouldn't work because these are extensions which are unsupported?

tafia:
So merged for now, I've opened an issue for extensions

Issue #92, opened 2018-02-12
Codegen: proto3 fields shouldn't have Option<> wrappers


scottlamb:
As you might know, proto3 drops the concepts of default values and even of a field being present or absent at all. You can see the [C++ API](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated) no longer has a `has_foo` accessor. The idea is that stuff can be represented with a simple struct.

I'd love for quick-protobuf to fully support this. In `tests/rust_protobuf/v3/test_basic_pb.proto`, this message:

```proto
message Test1 {
    int32 a = 1;
}
```

currently maps to this generated struct:

```rust
#[derive(Debug, Default, PartialEq, Clone)]
pub struct Test1 {
    pub a: Option<i32>,
}
```

the best proto3 representation leaves out that Option:

```rust
#[derive(Debug, Default, PartialEq, Clone)]
pub struct Test1 {
    pub a: i32,
}
```

with MessageRead and MessageWrite implementations simplified to match. A few reasons I prefer this representation:

   * it better matches intended semantics / improves interoperability. Any time you're using the equivalent of `has_*`, you're assigning a meaning that can't be represented in other languages, so you're losing portability.
   * it's simpler to deal with.
   * it's more memory-compact. IIUC, even with Rust's fancy field reordering, the whole Option has to be together, so I think it basically doubles the size of the in-memory representation.

fwiw, rust-protobuf drops the `Option<>` as I've suggested (but doesn't have `Cow` support, which is what led me to quick-rs).

```proto
#[derive(PartialEq,Clone,Default)]
pub struct Test1 {
    // message fields
    pub a: i32,
    // special fields
    unknown_fields: ::protobuf::UnknownFields,
    cached_size: ::protobuf::CachedSize,
}
```

(I also prefer having `unknown_fields`, but that's a separate issue.)

tafia:
Thanks for opening this issue and sorry for the late answer.

I think the best solution would be for pb-rs to have a dedicated flag so people can decide which solution best describes there files. And default to proto3 behaviour indeed.

Shouldn't be too hard. If you want to give it a try please do, else I will try to implement it on my side this month.

nerdrew:
I have a branch I've been testing for this. Question: should `Option<T>` for proto3 fields be configurable? I know that proto3 fields always have a default, but having a `Vec::new()` that is empty seems a little wasteful (though maybe not a big deal).

[...]

PR #129, opened 2018-10-23
proto3 scalar fields always have a default

nerdrew:
scalar fields should by type T instead of Option<T>: see
https://github.com/tafia/quick-protobuf/issues/92

### DIFF
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -56,7 +56,8 @@ impl MessageIndex {
             .skip(1)
             .fold(first_message, |cur, next| {
                 cur.and_then(|msg| msg.messages.get(*next))
-            }).expect("Message index not found")
+            })
+            .expect("Message index not found")
     }
 
     fn get_message_mut<'a>(&self, desc: &'a mut FileDescriptor) -> &'a mut Message {
@@ -69,7 +70,8 @@ impl MessageIndex {
             .skip(1)
             .fold(first_message, |cur, next| {
                 cur.and_then(|msg| msg.messages.get_mut(*next))
-            }).expect("Message index not found")
+            })
+            .expect("Message index not found")
     }
 
     fn push(&mut self, i: usize) {
@@ -549,7 +551,10 @@ impl Field {
                 w,
                 "if self.{} == {} {{ 0 }} else {{ {} + {} }}",
                 self.name,
-                self.default.as_ref().map_or_else(|| self.typ.regular_default(desc).unwrap_or("None"), |s| s.as_str()),
+                self.default.as_ref().map_or_else(
+                    || self.typ.regular_default(desc).unwrap_or("None"),
+                    |s| s.as_str()
+                ),
                 tag_size,
                 self.typ.get_size(&format!("&self.{}", self.name))
             )?,
@@ -628,7 +633,10 @@ impl Field {
                     w,
                     "        if self.{} != {} {{ w.write_with_tag({}, |w| w.{})?; }}",
                     self.name,
-                    self.default.as_ref().map_or_else(|| self.typ.regular_default(desc).unwrap_or("None"), |s| s.as_str()),
+                    self.default.as_ref().map_or_else(
+                        || self.typ.regular_default(desc).unwrap_or("None"),
+                        |s| s.as_str()
+                    ),
                     self.tag(),
                     self.typ
                         .get_write(&format!("&self.{}", self.name), self.boxed)
@@ -1685,7 +1693,8 @@ impl FileDescriptor {
                                     .filter(|f| f.frequency == Frequency::Required)
                                     .filter(|f| {
                                         f.typ.message().map_or(false, |m| cycle.contains(m))
-                                    }) {
+                                    })
+                                {
                                     f.boxed = true;
                                     f.frequency = Frequency::Optional;
                                 }
@@ -1771,7 +1780,8 @@ impl FileDescriptor {
                         vec![&mut **key, &mut **value].into_iter()
                     }
                     _ => vec![typ].into_iter(),
-                }) {
+                })
+            {
                 if let FieldType::MessageOrEnum(name) = typ.clone() {
                     let test_names: Vec<String> = if name.starts_with(".") {
                         vec![name.clone().split_off(1)]
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -18,7 +18,7 @@ fn sizeof_varint(v: u32) -> usize {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Syntax {
     Proto2,
     Proto3,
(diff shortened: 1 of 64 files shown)


tafia:
Thanks for this PR as well!
Can you run rustfmt stable here too?

tafia:
I'll probably take more time reviewing this one.

nerdrew:
Any thoughts on this PR?

tafia:
Sorry forgot about it!
Reviewing asap

pb-rs/src/types.rs
```diff
@@ -512,67 +545,103 @@ impl Field {
...
             }
+            Frequency::Optional => writeln!(
+                w,
+                "if self.{} == {} {{ 0 }} else {{ {} + {} }}",
+                self.name,
+                self.default
+                    .iter()
+                    .next()
+                    .map(|s| s.as_str())
+                    .unwrap_or(self.typ.regular_default(desc).unwrap_or("None")),
```
tafia:
```
self.default.as_ref().map_or_else(|| self.typ.regular_default(desc).unwrap_or("None"), s.as_str())
```

pb-rs/src/types.rs
```diff
@@ -512,67 +545,103 @@ impl Field {
...
+            Frequency::Optional => {
+                writeln!(
+                    w,
+                    "        if self.{} != {} {{ w.write_with_tag({}, |w| w.{})?; }}",
+                    self.name,
+                    self.default
+                        .iter()
+                        .next()
+                        .map(|s| s.as_str())
+                        .unwrap_or(self.typ.regular_default(desc).unwrap_or("None")),
```
tafia:
```
self.default.as_ref().map_or_else(|| self.typ.regular_default(desc).unwrap_or("None"), s.as_str())
```

pb-rs/src/types.rs
```diff
@@ -499,10 +507,35 @@ impl Field {
...
 
-    fn write_get_size<W: Write>(&self, w: &mut W) -> Result<()> {
+    fn write_get_size<W: Write>(&self, w: &mut W, desc: &FileDescriptor) -> Result<()> {
         write!(w, "        + ")?;
         let tag_size = sizeof_varint(self.tag());
         match self.frequency {
+            Frequency::Optional
+                if desc.syntax == Syntax::Proto2 || self.typ.message().is_some() =>
+            {
+                // TODO this might be incorrect behavior for proto2
```
tafia: can you clarify?

pb-rs/src/types.rs
```diff
@@ -1051,6 +1099,22 @@ impl Enumerator {
...
+            .map(|f| (format!("{}::{}", &self.name, f.0), f.1))
+            .collect();
+        self.fully_qualified_fields = self
+            .partially_qualified_fields
+            .iter()
+            .map(|pqf| {
+                let mut fqf = pqf.0.clone();
+                if self.module != "" {
+                    fqf.insert_str(0, format!("{}::", &self.module.replace(".", "::")).as_str());
+                }
```
tafia:
```
let fqf = if self.module.is_empty() {
    pqf.0.clone()
} else {
    format!("{}::{}", self.module.replace(".", "::"), pqf.0)
};
```

quick-protobuf/tests/rust_protobuf/common/mesos/v1.rs
```diff
@@ -1188,10 +1188,7 @@ pub struct HTTPCheckInfo<'a> {
 
 impl<'a> MessageRead<'a> for HTTPCheckInfo<'a> {
     fn from_reader(r: &mut BytesReader, bytes: &'a [u8]) -> Result<Self> {
-        let mut msg = HTTPCheckInfo {
-            protocol: mesos::v1::mod_NetworkInfo::Protocol::IPv4,
-            ..Self::default()
-        };
+        let mut msg = Self::default();
```
tafia: Looks suspicious.

quick-protobuf/tests/rust_protobuf/common/test_oneof.rs
```diff
@@ -1,8 +1,17 @@
 use quick_protobuf::*;
 
-use std::borrow::Cow;
-use rust_protobuf::hex::{encode_hex, decode_hex};
 use super::test_oneof_pb::*;
+use rust_protobuf::hex::{decode_hex, encode_hex};
+use std::borrow::Cow;
+
+fn t<F>(f: F)
```
tafia: Any reason to move it?

pb-rs/src/types.rs
nerdrew: Hmmm. I wrote this a while ago, so this is what I _think_ I meant:

For proto2, if a field is set, I believe it should be encoded on the wire, even if it is set to the default value. This looks like it sets the size to `0` if the set value matches the default (and I assume it then doesn't encode the default value on the wire). This issue is not new to this PR, I think I included the TODO accidentally.

pb-rs/src/types.rs
nerdrew: Did you mean (added a `|s|` to `s.as_str()`):

```rust
self.default.as_ref().map_or_else(|| self.typ.regular_default(desc).unwrap_or("None"), |s| s.as_str())
```

pb-rs/src/types.rs
nerdrew: Same question re: adding `|s|` as above.

quick-protobuf/tests/rust_protobuf/common/mesos/v1.rs
nerdrew: Looking at the generated code, I don't think it is functionally any different. The default value for `mesos::v1::mod_NetworkInfo::Protocol` is `IPv4`, so calling default on `HTTPCheckInfo` should use the default for `Protocol`, correct?

Thoughts?

quick-protobuf/tests/rust_protobuf/common/test_oneof.rs
nerdrew: I think I pulled it out of the method so I could use it in two different files since the behavior for v2 vs v3 is slightly different.

https://github.com/tafia/quick-protobuf/pull/129/files/99d1ebe96ff2eda118788b63454fea078ccd2d0f#diff-652110135f7c9e6e9bae3be74ac02869R5
https://github.com/tafia/quick-protobuf/pull/129/files/99d1ebe96ff2eda118788b63454fea078ccd2d0f#diff-693fe99aab3c7b2174edc30026c68f6fR6

nerdrew:
I rebased on master and force pushed the original commit + merge conflict resolution, then added a  `cargo +stable fmt` commit, and a PR feedback commit.

nerdrew:
Question: do you want me to include updated generated code in my PR (I did, but I can revert that bit if you don't)?

pb-rs/src/types.rs
tafia: yep

tafia:
Thanks a LOT and sorry for the delay in responses. I know it is not very nice to work like that.

PR #136, opened 2019-02-16
3 Changes to quick-protobuf

zutils:
1) Added the ability to use custom RPC generation code
2) Added the ability to modify custom derives such as derive-new
3) (Large change) Added the dont_use_cow method [v2 and v3 tested] which replaces lifetimes and Cow with regular strings and vectors. This will decrease performance - and will eventually be replaced with a whitelist for specific structures instead of a boolean.  This is meant for the structure to fully own the data it has.


zutils:
Hey Johann,
    For your question of why I use Ok(result?) instead of just result: I'm
used to using the failure crate. Sometimes just passing "result" back gives
me an error. I have fixed this error in the past by using Ok(result?). This
has become intuitive on my part. It is probably another unnecessary step
though! I'll keep that in mind for the future!

   For custom derive, I think passing in a map would be a good idea in a
future PR. I mostly use it for derive-new. There must be a better way of
doing this!



dont_use_cow: self.dont_use_cow, //Change this to false to not use cow
for v2 and v3 tests

     This is due to laziness on my part.  I didn't go around creating test
cases for dont_use_cow - instead, I re-used the existing ones you already
have.  If you change dont_use_cow to true at this line, then it will run
test cases using String instead of Cow.  Again - I want to whitelist
messages instead of making ALL of them not use Cow. If I can whitelist
specific cases, then I can create a few specific test case for dont_use_cow
instead.

I think I'd prefer a D but this is up to you :
    "D" is just fine.

It probably makes sense to use Arg::multiple
<https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.multiple> directly?
     This makes sense :)

Is it intentional to comment it?
    The comment can be taken out. Instead of using Cow::Borrow, I can use
.into()

And yes - convert_field_types can change.

I don't mind any of these changes "AT ALL". They are all very good. I can
do these changes within the next couple weeks.  If you get to them before
me - that's okay too.  Do you have any more thoughts?

V/R
Jim

tafia:
Awesome thank you!

`Ok(result?)` helps when you need to convert errors. I don't think that's the case here.

I probably won't have time to do it myself.

zutils:
Hi Johann,
    I just pushed these code changes up. I would like my changes to be in
the official quick-protobuf repository - so if there are any more changes
that you would like me to make, please let me know!
V/R
Jim

tafia:
Could you rebase to avoid the conflicts?

zutils:
Perhaps a bit of help may be needed. What I did is use the command "git
rebase". I also did fetches.  The message "Your branch is up to date with
'origin/master'." appeared. I do not see any conflicts. Do you know what
else I can do?

zutils:
Hi Johann,
Can you check up on this effort for me? I was able to rebase.  Thanks!
-Jim

tafia:
This is great thanks!
Could you just rust fmt it and I can merge it.
Thanks again for all your effort!

zutils:
Ok Done. No problem! Thanks for your help and cooperation!

tafia:
Thanks!
Just published pb-rs v0.8.0

PR #137, opened 2019-05-04
Use octal values for output of tag code values.

ZaneHannanAU:
Simpler reading of emitted source for values under 8 (and simpler reading of emitted type values regardless).

Mostly a readability thing for outputted code.

### DIFF
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -509,7 +509,7 @@ impl Field {
     fn write_match_tag<W: Write>(&self, w: &mut W, desc: &FileDescriptor) -> Result<()> {
         // special case for FieldType::Map: destructure tuple before inserting in HashMap
         if let FieldType::Map(ref key, ref value) = self.typ {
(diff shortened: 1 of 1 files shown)


tafia:
Hey, sorry for the late answer. May I know why do you think octal is better than hex for instance?
I don't really mind but to me it doesn't really improves anything. It could be even better not to show any tag value at all because it is of very little information for general consumers.

ZaneHannanAU:
In this case, it's because it displays somewhat better. For instance, rather than

```proto
message L {
  uint32 k = 1;
}
```

producing a (slightly nonsensical) `Ok(8)` value, it produces a (more readable) `Ok(0o10)` value.

 This expands to others, but does peter out after field 7 (`0o70`--`0o75`), where it becomes `0o100` for field 8; so obviously, this isn't perfect. It generates the same binary regardless (as in, it becomes collapsed) but it offers some insight for what types map to.

Other than that, literally no reason. Simply stronger patterns for humans to recognise.

PR #141, opened 2019-06-27
No need to create a Borrow to compare against a &str

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>


tafia:
Thanks!!

PR #147, opened 2019-09-09
rm generated files

nerdrew:


mullr:
It looks like a few generated files are still committed; is that intentional?

nerdrew:
Ya, @tafia asked to keep a few generated files so we can see what changes when the code gen changes (makes sense to me). I figured keeping the examples around would be helpful.

mullr:
Ah ok, cool. Looks good to me then.

tafia:
Thanks!!

PR #150, opened 2019-10-25
Replace some indexing by failable get

lexxvir:
Hi @tafia,

I've some trivial changes that prevents panics when unexpected EOF is reached.

### DIFF
--- a/quick-protobuf/src/reader.rs
+++ b/quick-protobuf/src/reader.rs
@@ -258,7 +258,12 @@ impl BytesReader {
     /// Reads fixed64 (little endian u64)
     #[inline]
     fn read_fixed<M, F: Fn(&[u8]) -> M>(&mut self, bytes: &[u8], len: usize, read: F) -> Result<M> {
-        let v = read(&bytes[self.start..self.start + len]);
+        let v = read(&bytes.get(self.start..self.start + len).ok_or_else(|| {
+            Error::Io(io::Error::new(
+                io::ErrorKind::UnexpectedEof,
+                "Cannot read next bytes",
+            ))
+        })?);
         self.start += len;
         Ok(v)
     }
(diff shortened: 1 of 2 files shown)


tafia:
Thanks!

quick-protobuf/src/reader.rs
```diff
@@ -338,14 +343,28 @@ impl BytesReader {
...
+        })
     }
 
     /// Reads string (String)
     #[inline]
     pub fn read_string<'a>(&mut self, bytes: &'a [u8]) -> Result<&'a str> {
         self.read_len_varint(bytes, |r, b| {
-            ::std::str::from_utf8(&b[r.start..r.end]).map_err(|e| e.into())
+            b.get(r.start..r.end)
+                .ok_or_else(|| {
```
nerdrew: Is there a performance benefit to skipping the `and_then` and using `?` instead? (Sorry for the very tardy response, just getting back into things... had a baby a bit ago, it's been hectic.)

Something like:

```rust
    /// Reads string (String)
    #[inline]
    pub fn read_string<'a>(&mut self, bytes: &'a [u8]) -> Result<&'a str> {
        self.read_len_varint(bytes, |r, b| {
            ::std::str::from_utf8(b.get(r.start..r.end).ok_or_else(|| {
                Error::Io(io::Error::new(
                    io::ErrorKind::UnexpectedEof,
                    "Cannot read next bytes",
                ))
            })?)
            .map_err(|e| e.into())
        })
    }
```

lexxvir: I think in both cases program's path will be the same:
 - try to get slice
 - if success then try to create `&str` from it

If `and_then()` called with `Err` variant it just returns this error.
Also, we talking about fail path, that IMHO, is very rare case in practice.

To be convinced we should benchmark both version on some real message handling.

tafia: I would be extremely surprised if they were any difference between the two.

nerdrew: Ya, I wasn't sure if closures allocated or not. It seems that they don't?

tafia: no they don't

PR #157, opened 2020-01-08
Added option to set `#[repr(...)]` on generated types.

ralfbiedert:
Hi, 

I added an option to specify a custom `#[repr(...)]`. We need this in our project since we also want to use the generated types as part of an FFI interface. 

PR includes a small bugfix that `custom_struct_derive` in the CLI wasn't annotated with `.takes_value(true)`.

Let me know if you want something changed!

### DIFF
--- a/pb-rs/src/lib.rs
+++ b/pb-rs/src/lib.rs
@@ -156,13 +156,13 @@ impl ConfigBuilder {
         self.custom_struct_derive = val;
         self
     }
-    
+
     /// Add custom values to `#[repr(...)]` at the beginning of every structure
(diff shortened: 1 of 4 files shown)


nerdrew:
I want to be able to add `#[serde(default)]` to a generated struct too. Should we keep adding flags for each case or should we have a general custom attribute flag?

ralfbiedert:
Good point. I think `-X "#[repr(C)]" -X "#[serde(default)]"` would actually be nicer.

tafia:
Merging, Thanks for the PR and sorry for the long delay

PR #158, opened 2020-01-23
Add no_std support

jpopesculian:
Haha let's try this a 3rd time... Building off of @xoloki's #148 work, I'd like to get this finalized and merged in. Let me know what you guys need to get this done


jpopesculian:
@nerdrew @tafia what's the status on this?

pb-rs/src/types.rs
```diff
@@ -312,7 +312,7 @@ impl FieldType {
                 format!("{}{}{}", m.get_modules(desc), m.name, lifetime)
             }
             FieldType::Map(ref key, ref value) => format!(
-                "HashMap<{}, {}>",
+                "KVMap<{}, {}>",
```
tafia: I believe most people won't use no_std and they'd rather use regular `std::HashMap`.
I understand that the distinction is done at import time but I'd like things to be as simple as possible for std users.

tafia:
```suggestion
                "{}<{}, {}>",
                    if config.no_std { "KVMap" } else { "HashMap" },
```

pb-rs/src/types.rs
```diff
@@ -1137,7 +1161,7 @@ impl Message {
     fn write_write_message<W: Write>(&self, w: &mut W, desc: &FileDescriptor) -> Result<()> {
         writeln!(
             w,
-            "    fn write_message<W: Write>(&self, w: &mut Writer<W>) -> Result<()> {{"
+            "    fn write_message<W: WriterBackend>(&self, w: &mut Writer<W>) -> Result<()> {{"
```
tafia: Where is the `WriterBackend` defined?

pb-rs/src/types.rs
jpopesculian: Hey @tafia, sorry for the delay in the reply! Thanks for taking a look :smile: I added the `KVMap` because the `rust_type` function doesn't have access to the config object (and I didn't want to change the call definition for a bunch of function), and the alias shouldn't restrict with what the actual person uses (other than reflecting in the docs). Obviously, I'd be happy to change it if you want, but I wanted to check with you first. I could of course also alias `BTreeMap` to `HashMap` but that seems a little hacky...

jpopesculian: I checked, the alias, actually does not reflect in the docs because its private. It's really pretty invisible to the user of the struct...

therealprof:
Can we please get this merged some time? 😅

tafia:
Sorry for the very long delay. Merging it once the CI tests pass.
Will deploy a new version just after

tafia:
Deployed v0.9.0 of pb-rs

therealprof:
Causes #165

PR #162, opened 2020-05-15
Attribute #![allow(clippy)] is deprecated, use #![allow(clippy::all)]…

therealprof:
… instead

Signed-off-by: Daniel Egger <daniel@eggers-club.de>


tafia:
Thanks

PR #171, opened 2020-07-27
Fix guard condition for out of bound access

lexxvir:
Hi @tafia,

It seems the condition should be fixed to actually guard out of bound access into the `self.buf`.


tafia:
Thanks!

PR #172, opened 2020-08-06
Remove unnecesarry `map_err`

lexxvir:
Hi @tafia,

It seems `map_err` already unnecessary because `WriterBackend` uses `Result` of the quick-protobuf.


tafia:
Thanks!

PR #174, opened 2020-08-11
Allow hexadecimal tag numbers in pb-rs

jpochyla:
Some of the 3rd party protobuf definitions I'm working with have the field tags specified in hex, the fix seems quick enough.

Thanks for the library!


tafia:
Thanks!

PR #175, opened 2020-08-11
Fix empty bytes field checking when --dont_use_cow flag is used.

### LINKED ISSUE #173: bug: Compile fails, inccorect code gen for 'bytes' with flag --dont_use_cow.
I illustrate with a sample:

for this message 
```
message Invoke {
  uint32 method = 1;
  uint64 action_id = 2;
  bool is_response = 3;
  bytes rpc_data = 4;
}
```

the output is:
```rust
pub struct Invoke {
    ...
    pub rpc_data: Vec<u8>,
}

impl MessageWrite for Invoke {
    fn get_size(&self) -> usize {
        ...
        + if self.rpc_data == vec![] { 0 } else { 1 + sizeof_len((&self.rpc_data).len()) }
    }

       ...
        if self.rpc_data != vec![] { w.write_with_tag(34, |w| w.write_bytes(&**&self.rpc_data))?; }
        Ok(())
    }
}
```

compiler fails to infer type:
```
error[E0282]: type annotations needed
   --> src/pb/pb/sys.rs:197:31
    |
197 |         + if self.rpc_data == vec![] { 0 } else { 1 + sizeof_len((&self.rpc_data).len()) }
    |                               ^^^^^^ cannot infer type for type parameter `T`
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
output should be something like:
```rust
+ if self.rpc_data.is_empty() { 0 } else { 1 + sizeof_len((&self.rpc_data).len()) }
```

jozn:
Closes #173.

### DIFF
--- a/pb-rs/src/types.rs
+++ b/pb-rs/src/types.rs
@@ -595,17 +595,26 @@ impl Field {
                     self.typ.get_size("")
                 )?;
             }
-            Frequency::Optional => writeln!(
-                w,
-                "if self.{} == {} {{ 0 }} else {{ {} + {} }}",
-                self.name,
-                self.default.as_ref().map_or_else(
-                    || self.typ.regular_default(desc).unwrap_or("None"),
-                    |s| s.as_str()
-                ),
-                tag_size,
-                self.typ.get_size(&format!("&self.{}", self.name))
-            )?,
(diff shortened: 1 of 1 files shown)


tafia:
Thanks!

PR #176, opened 2020-09-10
Add Error::Io for `no_std`

lexxvir:
Hi @tafia !

It will be nice if in the `no_std` environment I can throw IO error from the `WriterBackend`'s implementer.


tafia:
Thanks!

tafia:
Sorry I've merged but I am not sure it is doing anything just yet.

lexxvir:
I will use this error in my `WriterBackend` implementation. So now it is possible to"notify" caller of `Writer` that IO error occurred.

PR #177, opened 2020-09-10
Don't force inlining on `no_std`

lexxvir:
Hi @tafia !

I propose to don't force `inline` directive when `no_std` enabled. Forced code inline leads to huge stack sizes and when we have little memory in the embedded CPU it is matter.


tafia:
Sounds ok for me. I let this open in case someone has a different opinion.
I'll merge it next week.

PR #178, opened 2020-10-07
Fix update mod.rs for 'package a.b.c' case

lexxvir:
Hi @tafia !

I have encounter problem when use triple nested package such as

```protobuf
syntax = "proto2";

package a.b.c;

message ...
```

Root `mod.rs` file does not contain `pub mod a;` string.
If I use `package a.b;` all OK.

This PR works OK for both cases.
If you wish I may add a test for 'a.b.c' case.


tafia:
Thanks!

PR #180, opened 2020-11-13
Remove proc from keywords

semtexzv:
Remove proc from keywords. https://doc.rust-lang.org/reference/keywords.html


nerdrew:
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1270-2018-06-21
https://github.com/rust-lang/rust/pull/49699/

Seems good to me. Is there a documented "minimum supported rustc" for this crate?

tafia:
There is no MSRV documented. Might be good to add tough.

PR #181, opened 2020-11-13
Generate message metadata

semtexzv:
While working on [This lib](https://github.com/semtexzv/quix) I've encountered the need to get information about individual messages, This is the simplest way I could think of to get message identifiers.

Might be useful to someone else


tafia:
Thanks!
Could you put it behind a flag?

tafia:
Thanks!

PR #182, opened 2020-11-30
Fix master build

lexxvir:
Hi @tafia,

It seems last PRs have been broken the build of the master branch. This patch repairs it.


tafia:
Thanks!

PR #188, opened 2021-03-12
Ignore deprecated OneOf fields in the enum definition too

### LINKED ISSUE #187: Wrong code generated for OneOfs with deprecated fields
The generated Rust enum definitions for OneOfs with deprecated fields, like:
```
oneof data {
    bytes a = 1;
    bytes b = 2 [deprecated=true];
}
```
contains the deprecated fields, but the corresponding reader/writer code does not, causing a build error because of the missing `match` arms.

yzsolt:
Fixes #187.


tafia:
Thanks!

PR #191, opened 2021-03-18
Add a pb-rs option to generate deprecated fields

### LINKED ISSUE #189: Mark deprecated fields instead of ignoring them
Currently all deprecated Proto fields are ignored when generating Rust code, making them inaccessible. I think this isn't the right thing to do, they should be marked with `#[deprecated]` instead ([documentation](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html)).

If you agree, I'm up for making the necessary changes.

yzsolt:
Closes #189.


tafia:
Awesome! You've been really busy!

PR #199, opened 2021-11-08
Replaced nom 3.2 with 7. This replaces the entire parser module.

runfalk:
Nom 3 is a bit old now and I was curious enough to try my hands at modernizing the parser. I've tried to copy the old structure as much as possible but in some cases I simplified it/condensed some functions. All tests and the benchmarks seems to run fine, but this one likely requires a bit more thorough review than my previous uplift.

The one "breaking" change here is that the file is parsed as UTF-8 rather than bytes (but it seems like proto files must be utf-8 anyway https://github.com/protocolbuffers/protobuf/issues/1418).

### DIFF
--- a/pb-rs/Cargo.toml
+++ b/pb-rs/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/tafia/quick-protobuf"
 edition = "2018"
 
 [dependencies]
-nom = "3.2.1"
+nom = "7"
 log = "0.4.4"
 clap = { version = "2.33.1", optional = true }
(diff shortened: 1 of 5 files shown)


tafia:
Great
Thanks!

imotai:
@runfalk the new parser fails when parsing quick-protobuf/tests/rust_protobuf/common/mesos.proto 
```
./pb-rs ../../quick-protobuf/tests/rust_protobuf/common/mesos.proto 
pb-rs fatal error No message or enum were read;either definition might be invalid or there were only unsupported structures
```

PR #205, opened 2021-12-27
fix: .. the parser

FauxFaux:
The new parser has broken `./generate_modules.sh` in a number of places.

As a first ... pass, fix the parser to actually error on rejections, then fix all of the `.proto` files the parser just rejects, due to omissions or bugs.

 * single-line comment parser was accepting no comments, because `not` doesn't seem to work like that
 * missed whitespace handling for ` }` in e.g. `message`.
 * enums with multiple lines of comments between items
 * enums with trailing semis

### DIFF
--- a/pb-rs/src/parser.rs
+++ b/pb-rs/src/parser.rs
@@ -370,7 +370,7 @@ fn enumerator(input: &str) -> IResult<&str, Enumerator> {
             delimited(pair(tag("enum"), many1(br)), word, many0(br)),
             delimited(
                 pair(tag("{"), many0(br)),
-                separated_list0(br, enum_field),
+                separated_list0(many0(br), enum_field),
                 pair(many0(br), tag("}")),
             ),
         ),
@@ -617,6 +617,18 @@ mod test {
         } else {
             panic!("Could not parse reserved fields message");
         }
+    }
+
+    #[test]
+    fn enum_comments() {
+        let msg = r#"enum Turn {
+            UP = 1;
+            // for what?
+            // for what, you ask?
+            DOWN = 2;
+          }"#;
+        let en = assert_complete(enumerator(msg));
+        assert_eq!(2, en.fields.len());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


pb-rs/src/parser.rs
```diff
@@ -60,7 +60,7 @@ fn integer(input: &str) -> IResult<&str, i32> {
 fn comment(input: &str) -> IResult<&str, ()> {
     value(
         (),
-        tuple((tag("//"), many0(not(line_ending)), opt(line_ending))),
+        tuple((tag("//"), take_while(|c| !"\n\r".contains(c)), opt(line_ending))),
```
snproj, line 63: Just personal preference, but I would remove `opt(line_ending)`; would like to separate comment parsing from newline parsing.

pb-rs/src/parser.rs
```diff
@@ -327,7 +327,7 @@ fn message(input: &str) -> IResult<&str, Message> {
                 delimited(
                     tag("{"),
                     many0(delimited(many0(br), message_event, many0(br))),
-                    tag("}"),
+                    preceded(many0(br), tag("}")),
```
snproj, line 330: I'm not that familiar with `nom`, but I think `message_event()` already handles whitespace parsing internally, so I was thinking of using #217's solution to take advantage of that.

snproj (commented):
Hi! The PR looks good to me 😄 

The small change to comments parsing that I've proposed isn't major (literally removing that parameter), but it would affect the unit tests.

If you're okay with it, I already have a local branch which has your commits, and then makes that minor change and corresponding unit test adjustments (as well as a few other parser fixes). Once my org approves it, I can make a PR for it. If everything goes well, my plan would be to accept your PR first, and then apply my own PR on top.

If you want to adjust it yourself / have any strong opinions, let me know though! Either way, you'll be able to see/comment on my adjustments once the PR comes out. I'm just trying to push this particular thing through faster since without it a lot of other fixes are being blocked on my end.

PR #207, opened 2021-12-27
feat(pb-rs): --skip-write

FauxFaux:
An option to not generate `MessageWrite` implementations, which I don't need and don't want. Not important at all.

Follows #206 for no apparent reason.


tdudz:
would like this as well

is this project maintained anymore? :/

Issue #202, opened 2021-12-16
Serialization without prefix length


JalonWong:
Why do `serialize_into_slice()` and `deserialize_from_slice()` add an an extra length before the data? This seems to be incompatible with the official library.

Can I add some new functions like `serialize_into_slice_without_len()` and  `deserialize_from_slice_without_len()` ?

AlaaZorkane:
Yes! Not only `serialize...()` functions but also the `Writer`

AlaaZorkane:
Hey @JalonWong just letting you know that I opened a PR here: https://github.com/tafia/quick-protobuf/pull/208 - Please take a look and give your feedback if possible, thanks! 🙏

JalonWong:
> Hey @JalonWong just letting you know that I opened a PR here: #208 - Please take a look and give your feedback if possible, thanks! 🙏

The PR is pretty good.
Also, That will be more handy If the `serialize_into_slice*()` can return the encoded length.

AlaaZorkane:
> The PR is pretty good. Also, That will be more handy If the `serialize_into_slice*()` can return the encoded length.

Thanks! I added it with the appropriate tests in the PR, if there is anything else let me know on the PR thread 🙏

nerdrew:
FYI: You can serialize / deserialize protos without length prefixes like this:

```rust
let mut buf = Vec::new();
let mut writer = Writer::new(&mut buf);
MyMessage::default().write_message(&mut writer)?;

let mut reader = BytesReader::from_bytes(&buf);
let proto = MyMessage::from_reader(&mut reader, &buf)?;
```

snproj:
Hi! Since this seems to be a common stumbling block for new users, would you guys prefer:
1. Modify `serialize_into_slice()` and `deserialize_from_slice()` to no longer include the length prefix
2. Modify `serialize_into_slice()` and `deserialize_from_slice()` to no longer include the length prefix, AND add two new methods that keep the old behavior (still a breaking change)
3. Keep `serialize_into_slice()` and `deserialize_from_slice()` the same, but add documentation
4. Keep `serialize_into_slice()` and `deserialize_from_slice()` the same, but add documentation AND two new methods that follow expectations better (basically #208)

ScarboroughCoral:
I have the same issue when I want to use protocol buffers between rust and typescript, protobufjs decoded error because the prefix length.
I need a function without prefix length 😄

PR #208, opened 2021-12-30
Serialize/Deserialize functions without length prefix

AlaaZorkane:
Following: https://github.com/tafia/quick-protobuf/issues/202

This PR adds the following:

- [x] Message writer without prefixing a length
- [x] Slice serializer without prefixing a length
- [x] Reading messages that **aren't** prefixed with a length 
- [x] Reading slices that **aren't** prefixed with a length
- [x] Read/Write test for all the above implementations

### DIFF
--- a/quick-protobuf/src/lib.rs
+++ b/quick-protobuf/src/lib.rs
@@ -14,8 +14,10 @@ pub mod writer;
 
 pub use crate::errors::{Error, Result};
 pub use crate::message::{MessageInfo, MessageRead, MessageWrite};
-pub use crate::reader::{deserialize_from_slice, BytesReader};
-pub use crate::writer::{serialize_into_slice, BytesWriter, Writer, WriterBackend};
+pub use crate::reader::{deserialize_from_slice, deserialize_from_slice_without_len, BytesReader};
+pub use crate::writer::{
+    serialize_into_slice, serialize_into_slice_without_len, BytesWriter, Writer, WriterBackend,
+};
 
 #[cfg(feature = "std")]
 pub use crate::reader::Reader;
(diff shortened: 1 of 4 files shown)


snproj:
Hi! The PR looks good; I was actually considering a few options, since this seems to be a common stumbling block for new users:

- Modify `serialize_into_slice()` and `deserialize_from_slice()` to no longer include the length prefix
- Modify `serialize_into_slice()` and `deserialize_from_slice()` to no longer include the length prefix, AND add two new methods that keep the old behavior (still a breaking change)
- Keep `serialize_into_slice()` and `deserialize_from_slice()` the same, but add documentation
- Keep `serialize_into_slice()` and `deserialize_from_slice()` the same, but add documentation AND two new methods that follow expectations better (basically this PR)

What do you think? Something to consider is the fact that we might have breaking changes already anyway from other PRs #235 and #240.

ScarboroughCoral:
Is there any progress on this pr  :-)

snproj:
In my opinion, since we already have breaking changes from other PRs that look like they're going to be merged, let's go with option 1 above. That is, just change the functions to no longer include the prefix, and *don't* include any backwards-compatibility functions that do.

I don't know of any particular use for the functions that have length prefixes, so my instinct is to remove them for the sake of keeping the API intuitive and small. Are you okay with this @AlaaZorkane?

ScarboroughCoral:
I think can add two methods this pr or next pr
1. `serialize_into_slice_with_variant(variant: Option<T>, msg: M)` 
2. `deserialize_from_slice with_variant() -> Result<variant:Option<T>>`
because I don't know what kind of a binary message is, I can serialize the message with a variant at head, and deserialize the message getting the variant and check my variant table get the type of the binary message and deserialize it.

PR #209, opened 2022-02-01
Fix read unknown overflow

koivunej:
Found this overflow by fuzzing one of the crates using quick-protobuf. This fixes an error related to `BytesReader::read_unknown` that was the first fuzzing error I found.

My branch is based on 75a0517 which is not the most recent master, as the nom upgrade seems to have broken the testing.

I can follow up with a simple fuzzing target in the PR.

### DIFF
--- a/quick-protobuf/src/errors.rs
+++ b/quick-protobuf/src/errors.rs
@@ -26,6 +26,8 @@ pub enum Error {
     UnexpectedEndOfBuffer,
     /// The supplied output buffer is not large enough to serialize the message
     OutputBufferTooSmall,
+    /// Input requires larger than usize offsets
+    SizeOverflow,
 }
 
 /// A wrapper for `Result<T, Error>`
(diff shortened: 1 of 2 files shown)


koivunej:
Oh, I could had rebased this had you pinged me. Was there a maintainership issue I missed, could you point it out for me @snproj (assuming you are now a maintainer)?.

snproj:
Ahh so sorry @koivunej! I went around pinging some people but I think I missed you, my bad 😥 Not a maintainership issue at all, more PRs would be great.

I'm also new to open source, so to be honest I'm still learning the ropes on communication; do let me know if I'm doing anything strange/too fast/should ask first. Are you alright with the changes currently?

koivunej:
> I'm also new to open source, so to be honest I'm still learning the ropes on communication; do let me know if I'm doing anything strange/too fast/should ask first. Are you alright with the changes currently?

I think you are doing fine! I will try to get back to fuzzing this crate sometime in the future and I think the changes ended up looking ok. Apologies for the late reply, I did note your response but just forgot to respond. I guess that's the open source way :) Please feel free to ping me if I can be of assistance as a long time user of the crate, though not so knowledgeable of the protobuf.

PR #217, opened 2022-04-07
fix comment and whitespace parsing

nerdrew:


snproj (commented):
Hi! This PR looks good to me. I think #205 should handle the concerns I mentioned in the comments, so my tentative plan is to merge #205, and then this one, though it might need a bit of merge conflict fixing in between.

Locally, I've already got a branch with both #205's and your commits on it (authorship intact, of course!), so I've already done the conflict fixing, which I will hopefully submit a PR for soon.

Let me know if you want to do it yourself, though! Either way, I'll tag you once that PR is up.

PR #233, opened 2022-11-22
bump version numbers, update changelog, prep for release

snproj:
# Preparing for release
- `pb-rs` goes from `0.9.1` to `0.10.0` due to breaking change in #153 
- `quick-protobuf` goes from `0.8.0` to `0.8.1`, no breaking changes as far as I can tell
- no changes to `perftest`

PR #232, opened 2022-11-22
various improvements to Owned generated code

nerdrew:
impl TryFrom for lifetime-less messages

PR #234, opened 2022-11-22
update changelog

snproj:
should have been part of #233 but my mind slipped 😅

just added #232 since its an API change


snproj:
@nerdrew could you let me know if i've understood the PR correctly?

Issue #12, opened 2017-01-25
Not supported


tafia:
List of items not supported for the moment (that I found, it may not be exhaustive)
- [ ] import
- [ ] package
- [ ] option
- [ ] nested types
- [ ] map
- [ ] google common messages (empty etc ...)
- [ ] groups
- [ ] reserved field numbers
- [ ] extensions
- [ ] nested extensions
- [ ] oneofs
- [ ] service
- [ ] deprecated option
- [ ] aliases
- [ ] *any* message
- [ ] json conversion

kingxsp:
Do you have any plans for support json conversion? This is a very good feature！！！

tafia:
I have no plan for the short term (not much time) but I welcome any PR!
I believe this should be pretty easy to implement using serde or rustc_resialize.

Jragonmiris:
Yeah, I think if we solve #61 this will take care of itself.

dpryg:
#68 should be in this list explicitly.
proto3 support is very limited.

tafia:
> #68 should be in this list explicitly.

Done. What else do you think is missing for proto3?

brianjcj:
looking forwards for message option support!

kootenpv:
Please add serde to this list

mumbleskates:
suggestion: preserving unknown fields (e.g., read message with unknown fields, modify, write: new written message preserves unknown fields unchanged). many apis also allow detecting and explicitly dropping unknown fields.

0x26res:
Am I correct in saying you don't support [dynamic messages and reflection](https://docs.rs/protobuf/latest/protobuf/reflect/index.html)

Could it be added to the list? Though I think it is a lot of work.

harmic:
re: JSON support - I have found that you can transcode a protobuf message by adding `Serialize` to the derives for all structs and enums that were generated.

It will use Serde's default JSON mapping, which is not quite the same as the canonical Protobuf JSON representation. Wether that is important would depend on the use case.

Unfortunately the `--custom_struct_derive` option is not useful for this, because it only adds the derive to structs. For the time being I have resorted to:

```
sed -i 's/derive(Debug/derive(serde::Serialize,Debug/' filename.rs
```

Issue #235, opened 2022-11-23
API change to getters and setters


snproj:
# Proposal
I'd like to hear if there are any opinions on moving to a generated getter/setter API for fields in messages, enums etc. rather than simply making fields public and accessing them directly.

# Why?
<details>
<summary>Current implementation</summary>
<br>

> Currently, our generated structs simply expose all fields as public, and the user simply reads from and assigns to them directly.
>
> Default values are handled by setting struct fields to their default values during initialization, but there's no persistent concept of default values associated with fields after that. There is also no concept of field presence (i.e. which fields have been explicitly set, and which have not)

</details>

I'm trying to implement a more compliant Proto2/Proto3 distinction. This involves introducing field presence as well as remembering default values of a field beyond just initialization. It seems relatively awkward to hack in these concepts when there's no abstraction layer between the field value and the user where these concepts most directly live.

I'm not saying it's impossible to maintain the current interface while adding these concepts (I'm trying that out right now to see how it goes), but I think doing so might make it harder to maintain.

On a more minor point, I also feel that if I were a user, it would make things clearer. For example, if I have a Proto3 field `msg.x` with a default value of 0, setting that field to 0 should mean that it is not serialized. `set_x(0)` would make me more aware of these possible side effects than just `x = 0`.

lexxvir:
Hi @snproj !

I heavily use quick-protobuf in the embedded (no_std) project and like the approach with public fields in the generated structs. That approach is significantly less noisy than getters/setters one.

I compared Google's C++ generated code vs quick-protobuf.

Though I use only the Proto2.
For Proto3 (when all fields are optional) the setters/getters approach maybe more convenient.

nerdrew:
I find working with the struct fields very nice. The objects are just structs. I don't need to worry about a new API when passing the structs around. An added benefit is that I add `serde::Deserialize` to most of them, and It Just Works. I don't need to care that the struct came from a proto or JSON on the wire. My app treats them the same.

I hope we can come up with a way to make this work with the existing struct fields without resorting to setters.

snproj:
Thanks for the feedback! One concern I have can be shown in the following example:

```protobuf
syntax = "proto2";
message A {
    optional int32 a = 1 [default=3];
}
```

This generates some Rust struct with field `pub a: i32`. The value at initialization (under our current system) will be `3`.

[...]

PR #242, opened 2022-12-28
Parse extensions correctly. Allow options more.

sinistersnare:
the syntax `extend MESSAGENAME { ... }` is now legal, and parses to an `Extend` struct. The struct is not used in the compiler.

Also, options are now parsed and ignored in enums and messages.

such as:

```proto
message A {
  option an_option = "hello, world!";
  option (an_option_extension) = false;
}
```

This PR only involves the Parser. It does not compile the extend functionality, only allows it to compile.

I kind of went into this codebase blind to add this, so if I am doing something non-kosher, let me know! I tried not to step on any toes, but I couldn't help myself a couple times. If you want me to revert the couple of style changes, I will happily.

Also, can there be a new release? It seems there has not been one in > 1 year?

### DIFF
--- a/pb-rs/src/parser.rs
+++ b/pb-rs/src/parser.rs
@@ -2,8 +2,8 @@ use std::path::PathBuf;
 use std::str;
 
 use crate::types::{
-    Enumerator, Field, FieldType, FileDescriptor, Frequency, Message, OneOf,
-    RpcFunctionDeclaration, RpcService, Syntax,
+    Enumerator, Extend, Field, FieldType, FileDescriptor, Frequency, Message, OneOf,
+    RpcFunctionDeclaration, RpcService, Syntax, Extensions,
 };
 
 use nom::{
@@ -13,7 +13,7 @@ use nom::{
         alpha1, alphanumeric1, digit1, hex_digit1, multispace1, not_line_ending,
     },
     combinator::{map, map_res, opt, recognize, value, verify},
-    multi::{many0, many1, separated_list0, separated_list1},
+    multi::{many0, many1, separated_list1},
     sequence::{delimited, pair, preceded, separated_pair, terminated, tuple},
     IResult,
 };
@@ -26,6 +26,13 @@ enum MessageEvent {
     ReservedNums(Vec<i32>),
     ReservedNames(Vec<String>),
     OneOf(OneOf),
+    Extensions(Extensions),
+    Ignore,
+}
+
+#[derive(Debug, Clone)]
+enum EnumEvent {
+    Field((String, i32)),
     Ignore,
 }
 
@@ -37,6 +44,7 @@ enum Event {
     Message(Message),
     Enum(Enumerator),
     RpcService(RpcService),
+    Extend(Extend),
     Ignore,
 }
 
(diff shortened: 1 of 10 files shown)


pb-rs/src/parser.rs
```diff
@@ -400,29 +430,54 @@ fn enum_field(input: &str) -> IResult<&str, (String, i32)> {
...
+            }
+        }
+        enumerator
         },
     )(input)
 }
 
 fn option_ignore(input: &str) -> IResult<&str, ()> {
-    value((), delimited(tag("option"), take_until(";"), tag(";")))(input)
+    value((), delimited(pair(tag("option"), many1(br)), take_until(";"), tag(";")))(input)
```
sinistersnare, line 466: I changed this rule to ensure that this parser only matches `option ...` and not `optional ...`

pb-rs/src/parser.rs
```diff
@@ -823,8 +985,10 @@ mod test {
         }
     }
 
-    #[test]
-    fn test_missing_tokens() {
+    mod missing_tokens {
```
sinistersnare, line 988: I changed this from a function to an inner-module... This seems like a much better idea than having functions. This way, there is the organization, without the foot-gun of not calling functions!

pb-rs/src/types.rs
```diff
@@ -781,6 +781,16 @@ fn get_modules(module: &str, imported: bool, desc: &FileDescriptor) -> String {
...
 
+#[derive(Debug, Clone, Default)]
+pub struct Extend {
+    /// The message being extended.
+    pub name: String,
+    /// All fields that are being added to the extended message.
+    pub fields: Vec<Field>,
+}
+
+impl Extend {}
```
sinistersnare, line 792: I created this extend struct, which hopefully will be incorporated into the compiler eventually (possibly by me), but we may want to switch to #235 first.

pb-rs/src/types.rs
```diff
@@ -1476,6 +1485,22 @@ impl RpcService {
...
+    pub to: i32,
+}
+
+impl Extensions {
+    /// The max field number that can be used as an extension.
+    pub fn max() -> i32 {
+        536870911
+    }
+
+}
```
sinistersnare, line 1502: To be clear this is related, but orthogonal functionality to 

```
extend X { ... }
```

This is the functionality for 

```
message A {
  ...
  extensions 500 to max
}

message B {
  ...
  extensions 39 to 339
}
```

sinistersnare:
Would love some guidance on what is causing the test failures? They seem unrelated to my changes?

snproj:
Hi! Just took a quick peek at why it's failing, and from what I can tell it's just because the test cases for enum aliases weren't implemented before, and the test cases are hardcoded to assert that they aren't implemented (more precisely, it asserts that `pb-rs` should not be able to parse `quick-protobuf > tests > rust_protobuf > v2/3 > test_enum_alias_pb.proto`).

More precisely, in `quick-protobuf > tests > rust_protobuf`:

```bash
must_fail["v3/test_enum_alias_pb.proto"]="enum alias not implemented"
must_fail["v2/test_enum_alias_pb.proto"]="enum alias not implemented"
```

If they are in fact now implemented, then these should be removed.

snproj:
Oh there are new releases for [`pb-rs`](https://crates.io/crates/pb-rs) (0.10.0) and [`quick-protobuf`](https://crates.io/crates/quick-protobuf) (0.8.1), if that's what you meant

pb-rs/src/parser.rs
snproj: Nice catch!

sinistersnare:
I think Option 2 is the best choice. Currently, the only documentation for missing features I see is #12. Should we add some docs at the crate level about unsupported features?

snproj:
Hmm, I think that would make sense -- and then perhaps close #12 with a note that users should now reference the docs instead (to keep a single source of truth). This could be a different PR, I think.

For now, I reckon we could just remove the lines in `generate.sh` completely (not just comment them out), and also add a comment to `v2 + 3 > test_enum_alias_pb.proto` explaining that the test used to throw a parse error but no longer does?

sinistersnare:
I made the requested change (hopefully to your agreement). Making official docs and dropping #12 seems like a great idea!

PR #243, opened 2022-12-29
Add field presence

snproj:
Changes discussed in #235.

@lexxvir @nerdrew @thomaseizinger feel free to have a look see!

### DIFF
--- a/quick-protobuf/examples/pb_rs/data_types.rs
+++ b/quick-protobuf/examples/pb_rs/data_types.rs
@@ -49,9 +49,18 @@ impl<'a> From<&'a str> for FooEnum {
 }
 
 #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Debug, Default, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct BarMessage {
     pub b_required_int32: i32,
+}
+
+
+impl Default for BarMessage {
+    fn default() -> Self {
+        Self {
+            b_required_int32: 0i32,
+        }
+    }
 }
 
 impl<'a> MessageRead<'a> for BarMessage {
@@ -71,29 +80,29 @@ impl<'a> MessageRead<'a> for BarMessage {
 impl MessageWrite for BarMessage {
     fn get_size(&self) -> usize {
         0
-        + 1 + sizeof_varint(*(&self.b_required_int32) as u64)
+        + 1 + sizeof_varint((self.b_required_int32) as u64)
     }
 
     fn write_message<W: WriterBackend>(&self, w: &mut Writer<W>) -> Result<()> {
-        w.write_with_tag(8, |w| w.write_int32(*&self.b_required_int32))?;
+        w.write_with_tag(8, |w| w.write_int32(self.b_required_int32))?;
         Ok(())
     }
 }
 
 #[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Debug, Default, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct FooMessage<'a> {
     pub f_int32: Option<i32>,
     pub f_int64: Option<i64>,
     pub f_uint32: Option<u32>,
     pub f_uint64: Option<u64>,
     pub f_sint32: Option<i32>,
-    pub f_sint64: i64,
-    pub f_bool: bool,
+    pub f_sint64: Option<i64>,
+    pub f_bool: Option<bool>,
     pub f_FooEnum: Option<FooEnum>,
     pub f_fixed64: Option<u64>,
     pub f_sfixed64: Option<i64>,
-    pub f_fixed32: u32,
+    pub f_fixed32: Option<u32>,
     pub f_sfixed32: Option<i32>,
     pub f_double: Option<f64>,
     pub f_float: Option<f32>,
@@ -112,13 +121,49 @@ pub struct FooMessage<'a> {
     pub test_oneof: mod_FooMessage::OneOftest_oneof<'a>,
 }
 
+impl<'a> FooMessage<'a> {
+    pub const DEFAULT_F_SINT64: i64 = 4i64;
+    pub const DEFAULT_F_BOOL: bool = true;
+    pub const DEFAULT_F_FIXED32: u32 = 0u32;
+}
+
+impl<'a> Default for FooMessage<'a> {
+    fn default() -> Self {
+        Self {
+            f_int32: None,
+            f_int64: None,
+            f_uint32: None,
+            f_uint64: None,
+            f_sint32: None,
+            f_sint64: None,
+            f_bool: None,
+            f_FooEnum: None,
+            f_fixed64: None,
+            f_sfixed64: None,
+            f_fixed32: None,
+            f_sfixed32: None,
+            f_double: None,
+            f_float: None,
+            f_bytes: None,
+            f_string: None,
+            f_self_message: None,
+            f_bar_message: None,
+            f_repeated_int32: Vec::new(),
+            f_repeated_packed_int32: Vec::new(),
+            f_repeated_packed_float: Cow::from(Vec::new()),
+            f_imported: None,
+            f_baz: None,
+            f_nested: None,
+            f_nested_enum: None,
+            f_map: HashMap::new(),
+            test_oneof: mod_FooMessage::OneOftest_oneof::None,
+        }
+    }
+}
+
 impl<'a> MessageRead<'a> for FooMessage<'a> {
     fn from_reader(r: &mut BytesReader, bytes: &'a [u8]) -> Result<Self> {
-        let mut msg = FooMessage {
-            f_sint64: 4i64,
-            f_bool: true,
-            ..Self::default()
-        };
+        let mut msg = Self::default();
         while !r.is_eof() {
             match r.next_tag(bytes) {
                 Ok(8) => msg.f_int32 = Some(r.read_int32(bytes)?),
@@ -126,12 +171,12 @@ impl<'a> MessageRead<'a> for FooMessage<'a> {
                 Ok(24) => msg.f_uint32 = Some(r.read_uint32(bytes)?),
                 Ok(32) => msg.f_uint64 = Some(r.read_uint64(bytes)?),
                 Ok(40) => msg.f_sint32 = Some(r.read_sint32(bytes)?),
-                Ok(48) => msg.f_sint64 = r.read_sint64(bytes)?,
-                Ok(56) => msg.f_bool = r.read_bool(bytes)?,
+                Ok(48) => msg.f_sint64 = Some(r.read_sint64(bytes)?),
+                Ok(56) => msg.f_bool = Some(r.read_bool(bytes)?),
                 Ok(64) => msg.f_FooEnum = Some(r.read_enum(bytes)?),
                 Ok(73) => msg.f_fixed64 = Some(r.read_fixed64(bytes)?),
                 Ok(81) => msg.f_sfixed64 = Some(r.read_sfixed64(bytes)?),
-                Ok(93) => msg.f_fixed32 = r.read_fixed32(bytes)?,
+                Ok(93) => msg.f_fixed32 = Some(r.read_fixed32(bytes)?),
                 Ok(101) => msg.f_sfixed32 = Some(r.read_sfixed32(bytes)?),
                 Ok(105) => msg.f_double = Some(r.read_double(bytes)?),
                 Ok(117) => msg.f_float = Some(r.read_float(bytes)?),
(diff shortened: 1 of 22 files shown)


Fishrock123:
As discussed in https://github.com/tafia/quick-protobuf/issues/235#issuecomment-1372718617 this approach (`Option` directly) seems suitable for this library.

This PR coves my use case and solves my problems (at least any that I currently know about).

snproj:
oops misclick on the review request 😅 sorry nerdrew if you got any notification spam

my plan is to merge this in a few days to a week if there's no pushback; a review before then would be more ideal but I'll try and merge it regardless before too long so I can merge other things

quick-protobuf/examples/pb_rs/data_types.rs
```diff
@@ -164,71 +209,74 @@ impl<'a> MessageRead<'a> for FooMessage<'a> {
...
-        + self.f_sfixed32.as_ref().map_or(0, |_| 1 + 4)
-        + self.f_double.as_ref().map_or(0, |_| 1 + 8)
-        + self.f_float.as_ref().map_or(0, |_| 1 + 4)
-        + self.f_bytes.as_ref().map_or(0, |m| 1 + sizeof_len((m).len()))
-        + self.f_string.as_ref().map_or(0, |m| 2 + sizeof_len((m).len()))
-        + self.f_self_message.as_ref().map_or(0, |m| 2 + sizeof_len((m).get_size()))
-        + self.f_bar_message.as_ref().map_or(0, |m| 2 + sizeof_len((m).get_size()))
-        + self.f_repeated_int32.iter().map(|s| 2 + sizeof_varint(*(s) as u64)).sum::<usize>()
-        + if self.f_repeated_packed_int32.is_empty() { 0 } else { 2 + sizeof_len(self.f_repeated_packed_int32.iter().map(|s| sizeof_varint(*(s) as u64)).sum::<usize>()) }
+        + if self.f_int32.is_some() { 1 + sizeof_varint((*(self.f_int32.as_ref().unwrap())) as u64) } else { 0 }
```
thomaseizinger, line 212: The old code (using `map_or`) seems more idiomatic to me. Why change to this?

quick-protobuf/examples/pb_rs/data_types.rs
```diff
@@ -112,26 +121,62 @@ pub struct FooMessage<'a> {
     pub test_oneof: mod_FooMessage::OneOftest_oneof<'a>,
 }
 
+impl<'a> FooMessage<'a> {
+    pub const DEFAULT_F_SINT64: i64 = 4i64;
+    pub const DEFAULT_F_BOOL: bool = true;
+    pub const DEFAULT_F_FIXED32: u32 = 0u32;
```
thomaseizinger, line 127: As I mentioned in the issue, I think it is worth generating a getter for all fields that have a default value in the protobuf file.

If there are concerns about the generated code size, we can hide it behind a config flag so it can be turned off/on although I am not sure it is worth the complexity. If you have a proto file with explicit defaults, there is a good chance you will have to type the code to access it so we might as well generate it for the user.

quick-protobuf/examples/pb_rs/data_types.rs
```diff
@@ -49,11 +49,20 @@ impl<'a> From<&'a str> for FooEnum {
...
 }
 
+
+impl Default for BarMessage {
+    fn default() -> Self {
+        Self {
+            b_required_int32: 0i32,
+        }
+    }
+}
```
thomaseizinger, line 64: Is this change strictly necessary? I'd be nice if the generated code is as idiomatic as possible.

One reason we are moving to this library from protoc is because we will commit the generated code to Git. Makes the build process simpler, IDE integration is better and the code is actually decent to look at.

quick-protobuf/examples/pb_rs_v3/data_types.rs
```diff
@@ -168,75 +210,78 @@ impl<'a> MessageRead<'a> for FooMessage<'a> {
...
+            mod_FooMessage::OneOftest_oneof::f1(m) => 2 + sizeof_varint((*m) as u64),
+            mod_FooMessage::OneOftest_oneof::f2(m) => 2 + sizeof_varint((*m) as u64),
             mod_FooMessage::OneOftest_oneof::f3(ref m) => 2 + sizeof_len((m).len()),
             mod_FooMessage::OneOftest_oneof::None => 0,
-    }    }
+        }
+    }
 
     fn write_message<W: WriterBackend>(&self, w: &mut Writer<W>) -> Result<()> {
-        if self.f_int32 != 0i32 { w.write_with_tag(8, |w| w.write_int32(*&self.f_int32))?; }
```
thomaseizinger, line 207: The diff in the generated code would be easier to review if this wouldn't be part of this PR :)

quick-protobuf/examples/pb_rs_v3/data_types.proto
```diff
@@ -17,12 +17,12 @@ message FooMessage {
...
     sint32 f_sint32 = 5;
-    sint64 f_sint64 = 6 [default=4];
-    bool f_bool = 7 [default=true];
+    sint64 f_sint64 = 6;
+    bool f_bool = 7;
     FooEnum f_FooEnum = 8;
     fixed64 f_fixed64 = 9;
     sfixed64 f_sfixed64 = 10;
-    fixed32 f_fixed32 = 11 [default=0];
+    fixed32 f_fixed32 = 11;
```
thomaseizinger, line 25: Are the changes to these test files strictly necessary?

quick-protobuf/examples/pb_rs_v3/data_types.rs
```diff
@@ -168,75 +210,78 @@ impl<'a> MessageRead<'a> for FooMessage<'a> {
...
-        + if self.f_sfixed32 == 0i32 { 0 } else { 1 + 4 }
-        + if self.f_double == 0f64 { 0 } else { 1 + 8 }
-        + if self.f_float == 0f32 { 0 } else { 1 + 4 }
-        + if self.f_bytes == Cow::Borrowed(b"") { 0 } else { 1 + sizeof_len((&self.f_bytes).len()) }
-        + if self.f_string == "" { 0 } else { 2 + sizeof_len((&self.f_string).len()) }
-        + self.f_self_message.as_ref().map_or(0, |m| 2 + sizeof_len((m).get_size()))
-        + self.f_bar_message.as_ref().map_or(0, |m| 2 + sizeof_len((m).get_size()))
-        + if self.f_repeated_int32.is_empty() { 0 } else { 2 + sizeof_len(self.f_repeated_int32.iter().map(|s| sizeof_varint(*(s) as u64)).sum::<usize>()) }
-        + if self.f_repeated_packed_int32.is_empty() { 0 } else { 2 + sizeof_len(self.f_repeated_packed_int32.iter().map(|s| sizeof_varint(*(s) as u64)).sum::<usize>()) }
+        + if self.f_int32 != 0i32 { 1 + sizeof_varint((self.f_int32) as u64) } else { 0 }
```
thomaseizinger, line 213: Same here, this change in the generated code seems unnecessary :)

thomaseizinger (commented):
Left some comments. Overall looks good to me. I wonder how we can best document what `Default::default` will do but I think this implementation is better than the one before :)

thomaseizinger:
@snproj Are you still working on this? We currently migrating to `quick-protobuf` but we definitely need this branch before we can roll that out to handle `proto3` correctly.

thomaseizinger:
Friendly ping @snproj . Let me know if there is anything I can help with to push this over the finish line, we are quite interested in having these changes land!

snproj:
Hello! So, so sorry, stuff cropped up on my end and I took a bit of a long break 😅 I'm back now, and will generally go through everything I've missed out over the past weeks.

thomaseizinger:
> Hello! So, so sorry, stuff cropped up on my end and I took a bit of a long break 😅 I'm back now, and will generally go through everything I've missed out over the past weeks.

No worries at all! Like I said, happy to take over any work here as well if you want! We are interested in having this land in for `rust-libp2p`.

quick-protobuf/examples/pb_rs/data_types.rs
snproj: Point taken; have changed this back to `map_or()`

The change away was originally just a side effect of me trying to make the codegen less fiddly for maintenance purposes -- have since realized I can implement `map_or()` without too much mess.

quick-protobuf/examples/pb_rs/data_types.rs
snproj: Yep, working on this currently -- this is the main holdup as the codegen is getting a bit fiddly

quick-protobuf/examples/pb_rs/data_types.rs
snproj: Made a change that will avoid custom `impl Default` on structs that don't need it; should be present in next update

quick-protobuf/examples/pb_rs_v3/data_types.rs
snproj: Yeah I think my rewriting some of the codegen led to me changing some details without realizing; let me check that the original version still plays nicely with the rest of the generated code. If so this will probably come under a tidy-up PR.

quick-protobuf/examples/pb_rs_v3/data_types.proto
snproj: That file contains Proto3, which doesn't allow for custom defaults from what I know, hence I removed those

quick-protobuf/examples/pb_rs_v3/data_types.rs
snproj: Whoop, yeah I think this was just changed by accident from the original as I rewrote some parts of the function; I'll change it back

snproj:
Just a quick update, I've made most of the changes above; the main thing left to do is the getters and setters, because codegen is a bit fiddly haha. Hope to finish that by this weekend though.

Also, I no longer have access to the `ghpr-asia` fork that this PR originally came from (I was originally interning at their company), so I would need to open a new PR from my personal fork for the changes, which I will link to this PR (and probably close this one at that point).

quick-protobuf/examples/pb_rs_v3/data_types.proto
thomaseizinger: Oh interesting, so you are saying that this proto file is invalid? Might be worth sending in a different PR if it is not too much hassle. (For bisect reasons)

snproj:
Just really quickly, as mentioned in the review

>If you have a proto file with explicit defaults, there is a good chance you will have to type the code to access it so we might as well generate it for the user.

I realized that the getters will mainly be used for fields with explicit custom defaults, so I think we can generate getters for just those, at least for now.

I was trying to generate for fields with *non-custom* defaults too (so Proto2 with no explicit default tag, and Proto3), but things like `Cow<str>` cause issues: A getter for a field of type `Cow<str>` would either return
1. `&Cow<str>`; but this means I can't return new values like `Cow::<str>::default()` (reference to temp value)
2.  `Cow<str>`; but this means I have to clone all the contents
3. `&str`, but that seems a bit weird

In the interest of time, I'll most likely work on generating for custom default fields only (i.e. Proto2 with explicit default tag) for this PR and we can expand later if needed.

snproj:
Just really quickly, as mentioned in the review

> If you have a proto file with explicit defaults, there is a good chance you will have to type the code to access it so we might as well generate it for the user.

I realized that the getters will mainly be used for fields with explicit custom defaults, so I think we can generate getters for just those, at least for now.

I was trying to generate for fields with *non-custom* defaults too, but things like `Cow<str>` cause issues: I would think a getter for a field of type `Cow<str>` should return `&Cow<str>`; but if I want to take standard default `Cow::<str>::default()` into account it's not possible (returning reference to newly created value).

<details>
<summary> Other options I don't think are ideal </summary>

1. Return `Cow::<str>`: Need to clone data
2. Return `&str`: possible, but a bit unintuitive IMO
3. Create global `const DEFAULT_cow_str: Cow::<str> = Cow::Borrowed("");` and return a reference to that

</details>

In the interest of time, I'll most likely work on generating for custom default fields only (i.e. Proto2 with explicit default tag) for this PR and we can expand later if needed.

thomaseizinger:
> > If you have a proto file with explicit defaults, there is a good chance you will have to type the code to access it so we might as well generate it for the user.
> 
> I realized that the getters will mainly be used for fields with explicit custom defaults, so I think we can generate getters for just those, at least for now.

Yeah, I never expected getters for the other fields? If there is no explicit default defined, you can just access the field without any surprises right? Perhaps what would be good is to add some documentation to those fields, such that when you look at the generated code, you are reminded that for those you should use the getter to make sure the default is taken into account.

quick-protobuf/examples/pb_rs_v3/data_types.proto
snproj: Yep, I've separated these out for a new PR after this one (well, technically #247) goes through!

snproj:
I've opened the new PR #247 with updates as a continuation since I no longer have access to the `ghpr-asia` fork; let's continue discussion there!

PR #247, opened 2023-02-27
Field presence

snproj:
# Continuation of #243
Since I no longer have access to the `ghpr-asia` fork that I made that PR with.

- Some minor formatting changes rolled back to reduce the diff
- Custom default `impl`s should now only be generated when there are custom default fields
- Added `--generate-getters` to generate getters for Proto2 fields with custom defaults

## Minor notes:
- Forgot to mention in previous PR message, but custom default fields for strings and bytes must now begin and end with either single or double inverted commas (previously not enforced). I had to enforce this to predictably generate default `const` fields from user input.
- The minor change from `#[derive(Debug, Default, PartialEq, Clone)]` to `#[derive(Default, Debug, PartialEq, Clone)]` is there because `Default` is now being inserted programmatically into the list of derives, so it was best to simply append to the start

## Future action
- Proto2 and Proto3 parsing will be made stricter; originally this was part of this PR (since the whole parsing system got restructured anyway), but it's probably better to separate it out. The changes are:
    - Proto3 enums must begin with index `0`
    - Proto3 cannot have `required` fields
    - Proto3 cannot have custom defaults


quick-protobuf/examples/pb_rs/data_types.rs
```diff
@@ -112,26 +112,63 @@ pub struct FooMessage<'a> {
...
+            f_repeated_packed_float: PackedFixed::from(Vec::new()),
+            f_imported: None,
+            f_baz: None,
+            f_nested: None,
+            f_nested_enum: None,
+            f_map: HashMap::new(),
+            test_oneof: mod_FooMessage::OneOftest_oneof::None,
+        }
+    }
+}
```
thomaseizinger, line 154: Why is this one not derived?

quick-protobuf/examples/pb_rs_nostd/protos/no_std.rs
```diff
@@ -114,15 +114,15 @@ impl<'a> MessageWrite for NoStdMessage<'a> {
...
     }
 
     fn write_message<W: WriterBackend>(&self, w: &mut Writer<W>) -> Result<()> {
         if self.num != 0u32 { w.write_with_tag(13, |w| w.write_fixed32(*&self.num))?; }
-        w.write_packed_fixed_with_tag(18, &self.nums)?;
-        if let Some(ref s) = self.message { w.write_with_tag(26, |w| w.write_message(s))?; }
-        for s in &self.messages { w.write_with_tag(34, |w| w.write_message(s))?; }
+                w.write_packed_fixed_with_tag(18, &self.nums)?;
+        w.write_with_tag(26, |w| w.write_message(&self.message))?;
+                for s in &self.messages { w.write_with_tag(34, |w| w.write_message(s))?; }
```
thomaseizinger, line 125: Unnecessary indentation?

quick-protobuf/examples/pb_rs_v3/data_types.rs
```diff
@@ -168,75 +164,78 @@ impl<'a> MessageRead<'a> for FooMessage<'a> {
...
-        match self.test_oneof {            mod_FooMessage::OneOftest_oneof::f1(ref m) => { w.write_with_tag(216, |w| w.write_int32(*m))? },
-            mod_FooMessage::OneOftest_oneof::f2(ref m) => { w.write_with_tag(224, |w| w.write_bool(*m))? },
-            mod_FooMessage::OneOftest_oneof::f3(ref m) => { w.write_with_tag(234, |w| w.write_string(&**m))? },
+        if self.f_bytes != Cow::Borrowed(b"") { w.write_with_tag(122, |w| w.write_bytes(&self.f_bytes))?; }
+        if self.f_string != Cow::Borrowed("") { w.write_with_tag(130, |w| w.write_string(&self.f_string))?; }
+        self.f_self_message.as_ref().map_or(Ok(()), |m| w.write_with_tag(138, |w| w.write_message(&*(*(m)))))?;
+        w.write_with_tag(146, |w| w.write_message(&self.f_bar_message))?;
+                w.write_packed_with_tag(154, &self.f_repeated_int32, |w, &m| w.write_int32(*&m), &|&m| sizeof_varint((m) as u64))?;
+                w.write_packed_with_tag(162, &self.f_repeated_packed_int32, |w, &m| w.write_int32(*&m), &|&m| sizeof_varint((m) as u64))?;
+                w.write_packed_fixed_with_tag(170, &self.f_repeated_packed_float)?;
```
thomaseizinger, line 224: Here too? Can we avoid this indentation?

quick-protobuf/tests/rust_protobuf/v2/test_default_values_test.rs
```diff
@@ -1,44 +1,122 @@
...
-    assert_eq!(9, d.fixed32_field);
-    assert_eq!(10, d.fixed64_field);
-    assert_eq!(11, d.sfixed32_field);
-    assert_eq!(12, d.sfixed64_field);
-    assert_eq!(true, d.bool_field);
-    assert_eq!("abc\n22", d.string_field);
-    assert_eq!(b"cde\n33".to_vec(), d.bytes_field.to_vec());
-    assert!(EnumForDefaultValue::TWO.eq(&d.enum_field));
+    let d = TestDefaultValuesOptional::from_reader(&mut reader, bytes).unwrap();
+    assert_eq!(1.0, d.double_field.unwrap_or(TestDefaultValuesOptional::DEFAULT_double_field));
```
thomaseizinger, line 14: Shouldn't this assert be that it is `None` and a separate assert that `DEFAULT_double_field == 1.0`?

thomaseizinger (commented):
Nice work!

I only reviewed the generated code as I don't have any knowledge on how the parsing system works.

snproj:
Made the changes as per above, also removed some extra parentheses in the generated `get_size()`es that previous versions of this PR added in.

quick-protobuf/examples/pb_rs/data_types.rs
snproj: Oversight in the `impl Default` system; have fixed it😅 thanks for catching

quick-protobuf/tests/rust_protobuf/v2/test_default_values_test.rs
snproj: Yep, I think I originally wanted to test exactly what the user would write (i.e. `obj.field.unwrap_or(Class::DEFAULT_field)` so I just wrote it in verbatim

Have changed it though since your version makes more sense haha

snproj:
I've spent a few hours trying to get the diff as small as possible for the `get_size()` and `write_message()` parts, but the way that the code generation has changed in this PR means that eliminating all the differences is hard without a bunch of error-prone rewriting (that will eventually be reverted anyway when I make the PR to clean it up, since I already have the code for that). I don't know that it's impossible, but I didn't want to delay too long haha

If you're curious, the previous code had control flow in various parts that determine when to add a dereference into the code, which had built up over time into generating a bunch of `&&*&**&`. This is the main source of diff, since I decided to rewrite the control flow to avoid these chains of references and dereferences; which makes it hard to artificially add them back in and shrink the diff.

Hopefully it's ok 🙏

thomaseizinger (commented):
LGTM! (Only reviewed generated code)

I'll ping a PR on our repo that needs these changes and have them try them out!

thomaseizinger:
> I've spent a few hours trying to get the diff as small as possible for the `get_size()` and `write_message()` parts, but the way that the code generation has changed in this PR means that eliminating all the differences is hard without a bunch of error-prone rewriting (that will eventually be reverted anyway when I make the PR to clean it up, since I already have the code for that). I don't know that it's impossible, but I didn't want to delay too long haha
> 
> If you're curious, the previous code had control flow in various parts that determine when to add a dereference into the code, which had built up over time into generating a bunch of `&&*&**&`. This is the main source of diff, since I decided to rewrite the control flow to avoid these chains of references and dereferences; which makes it hard to artificially add them back in and shrink the diff.
> 
> Hopefully it's ok pray

Yeah that is fine. There is a means to an end when it comes to reducing diff :)

thomaseizinger:
https://github.com/libp2p/rust-libp2p/pull/3455 uses this branch and it works well, green light from our end!

d-roak:
Hi @snproj!

Can you do a version bump in the near future to include this change?

Thanks 🙏

mxinden:
Friendly ping. Could you cut a new release with this patch?

PR #257, opened 2024-01-25
perftest: missing field in `Config`

thog92:
Fixes compile error due to missing field (`generate_getters`, introduced in #247).


dheeraj-grasshopper:
lgtm

